PR #189, opened 2019-07-01
router can apply multi router with same url and different http method

DemonsLu:
解决issue #188 的问题，将原生的http.Mux换成github.com/gorilla/mux


tonyboxes:
已经支持restful

PR #240, opened 2019-07-25
refine main.go template

HAOYUatHZ:
refine main.go template following the style in https://gobyexample.com/range-over-channels


ChrisLinn:
@DeKaiju

PR #385, opened 2019-10-15
support get the actual listened address after grpc server with zero port is started

wangrzneu:
when the configured port is zero, the os will allocate an unused random port to the server, the actual port needs to be got for service registry


Terry-Mao:
@tonyboxes

wangrzneu:
@tonyboxes Please help to review this change. Thanks!

PR #393, opened 2019-10-23
multi zone scheduler sdk

lintanghui:


lintanghui:
@tonyboxes @longXboy PTAL

PR #430, opened 2019-11-17
opt last append time of rolling policy

mikellxy:
There would be a problem if update the lastAppendTime to time.Now() after a token was filled into a bucket. Specifically, assuming the bucket duration is 300ms, if I call the MarkSuccess at 400ms and 601ms, what we expect is that the second and third bucket will be filled with one token respectively, however, there will be two tokens filled into the second bucket.  

Following is my test code:  
```go
package kratos

import (
	"github.com/bilibili/kratos/pkg/net/netutil/breaker"
	"testing"
	"time"
)

func TestBreaker(t *testing.T) {
	brkGroup := breaker.NewGroup(&breaker.Config{})
	brk := brkGroup.Get("name")

	time.Sleep(400 * time.Millisecond)
	brk.MarkSuccess()
	time.Sleep(201 * time.Millisecond)
	brk.MarkSuccess()

}
```
Print the bucket points after each MarkSuccess:  
```go
func (r *RollingPolicy) add(f func(offset int, val float64), val float64) {
	r.mu.Lock()
	timespan := r.timespan()
	if timespan > 0 {
		offset := r.offset
		// reset the expired buckets
		s := offset + 1
		e, e1 := s+timespan, 0 // e: reset offset must start from offset+1
		if e > r.size {
			e1 = e - r.size
			e = r.size
		}
		for i := s; i < e; i++ {
			r.window.ResetBucket(i)
			offset = i
		}
		for i := 0; i < e1; i++ {
			r.window.ResetBucket(i)
			offset = i
		}
		r.offset = offset
		r.lastAppendTime = time.Now()
	}
	f(r.offset, val)
	r.mu.Unlock()

	fmt.Println("*****buckets*****")
	for _, b := range r.window.window {
		fmt.Println(b.Points)
	}
}
```

The result is:  
```go
=== RUN   TestBreaker
*****buckets*****
[]
[1]
[]
[]
[]
[]
[]
[]
[]
[]
*****buckets*****
[]
[2]
[]
[]
[]
[]
[]
[]
[]
[]
```

### DIFF
--- a/pkg/stat/metric/rolling_policy.go
+++ b/pkg/stat/metric/rolling_policy.go
@@ -38,7 +38,7 @@ func NewRollingPolicy(window *Window, opts RollingPolicyOpts) *RollingPolicy {
 
 func (r *RollingPolicy) timespan() int {
 	v := int(time.Since(r.lastAppendTime) / r.bucketDuration)
-	if v < r.size && v > -1 { // maybe time backwards
+	if v > -1 { // maybe time backwards
 		return v
 	}
 	return r.size
@@ -48,9 +48,13 @@ func (r *RollingPolicy) add(f func(offset int, val float64), val float64) {
 	r.mu.Lock()
 	timespan := r.timespan()
 	if timespan > 0 {
+		r.lastAppendTime = r.lastAppendTime.Add(time.Duration(timespan * int(r.bucketDuration)))
 		offset := r.offset
 		// reset the expired buckets
 		s := offset + 1
+		if timespan > r.size {
+			timespan = r.size
+		}
 		e, e1 := s+timespan, 0 // e: reset offset must start from offset+1
 		if e > r.size {
 			e1 = e - r.size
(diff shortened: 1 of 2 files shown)


Terry-Mao:
lgtm

pkg/stat/metric/rolling_policy.go
```diff
@@ -65,7 +65,7 @@ func (r *RollingPolicy) add(f func(offset int, val float64), val float64) {
 			offset = i
 		}
 		r.offset = offset
-		r.lastAppendTime = time.Now()
+		r.lastAppendTime = r.lastAppendTime.Add(time.Duration(timespan * int(r.bucketDuration)))
```
faycheng: if time.since(r.lastAppendTime) > r.size * r.bucketDuration, r.timespan returns r.size, not real time span.
maybe we should replace timespan with `time.since(r.lastAppendTime) / r.bucketDuration`.

faycheng:
Thanks a lot for this pr. Can we write some test cases to verify it?

pkg/stat/metric/rolling_policy.go
faycheng: another solution, returns the real time span in func `timespan`, and changes the logic of calculating the variable `e`.

mikellxy: I prefer that func `timespan` returns the real time span, because it is more accurate to use `time.Since` once when `RollingPolicy.Add` called. Some test cases were supplied. Thanks for your review~

faycheng:
LGTM

PR #572, opened 2020-04-24
Refine/error handling

F1renze:
- provide an easy way to handling error across service boundaries

before:

```go
// Server end
func (s *rpcServer) Call(ctx context.Context, req *pb.Request) (*pb.Response, error) {
	if len(req.Name) < 1 {
		return nil, ecode.Error(ecode.RequestErr, "name is required")
	}
	return &pb.Response{Status: 1}, nil
}
```

```go
import (
	...
	st "github.com/go-kratos/kratos/pkg/net/rpc/warden/internal/status"
	"github.com/go-kratos/kratos/pkg/ecode/types"
	...
)

...

// Client end
reply, err := rpcCli.Call(ctx, req)
// got grpc.Status
gst := st.FromError(err)
if len(gst.Details()) > 0 {
	if st, ok := gst.Details()[0].(*types.Status); ok {
		// Use kratos ecode.Status.Code
	}
}
//  convert grpc Status to kratos ecode.Codes
ec := st.ToMetaCode(gst)
// Use ec.Code(), ec.Message()
```

now:
```go
// Client end
reply, err := rpcCli.Call(ctx, req)
if err != nil {
	// convert error to ecode.Status directly
	st := ecode.Parse(err)
	if st.Code() == ecode.RequestErr.Code() {
		// deal with bad request
	}
}
```

### DIFF
--- a/pkg/ecode/status.go
+++ b/pkg/ecode/status.go
@@ -1,6 +1,7 @@
 package ecode
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strconv"
@@ -9,6 +10,8 @@ import (
 
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc/status"
 )
 
 // Error new status with code and message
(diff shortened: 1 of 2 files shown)


F1renze:
PTAL

PR #589, opened 2020-05-20
feat: add bm context Value().

### LINKED ISSUE #588: bm context 是否也可以支持一下Value() 方法？
如题，Gin也是支持的，见[代码](https://github.com/gin-gonic/gin/blob/1d5b9ba/context.go#L1101).

Colstuwjx:
This PR is going to fix #588.

### DIFF
--- a/pkg/net/http/blademaster/context.go
+++ b/pkg/net/http/blademaster/context.go
@@ -367,6 +367,20 @@ func (c *Context) Bind(obj interface{}) error {
 	return c.mustBindWith(obj, b)
 }
 
+// Value returns the value associated with this context for key, or nil
+// if no value is associated with key. Successive calls to Value with
+// the same key returns the same result.
+func (c *Context) Value(key interface{}) interface{} {
+	if key == 0 {
+		return c.Request
+	}
+	if keyAsString, ok := key.(string); ok {
+		val, _ := c.Get(keyAsString)
+		return val
+	}
+	return nil
+}
+
 // mustBindWith binds the passed struct pointer using the specified binding engine.
 // It will abort the request with HTTP 400 if any error ocurrs.
 // See the binding package.


Colstuwjx:
Just like Gin did, I think bm also need this method, therefore, we could make `bm.Context` compatible with the golang Context interface.

Colstuwjx:
I'm not sure why the test case run failed:

```
--- FAIL: TestCriticality (1.00s)
    server_test.go:96: 
        	Error Trace:	server_test.go:96
        	Error:      	Not equal: 
        	            	expected: "CRITICAL"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(criticality.Criticality) (len=8) "CRITICAL"
        	            	+(criticality.Criticality) ""
        	            	 
        	Test:       	TestCriticality
```

realityone:
This will break the derived context using `Value` method to retrieve the value like deadline or timeout.

As test case is failed, `Criticality` is stored in `valueContext`, not `bm.Context.Keys`.

Colstuwjx:
Little confused to me.

so after pass the bm context into a new svc function by `context.Context` interface, we could ONLY touch the [inner context](https://github.com/go-kratos/kratos/blob/master/pkg/net/http/blademaster/context.go#L36)(this is so-called valueContext?) values, such as metadata?

My question is how to get the `bm.Context.Keys` while we're inside svc function with passed context.Context as input?

realityone:
Using `bm.Context` in service is not recommended.

I'd prefer to prepare your arguments from `bm.Context`  in http package and never touch this in any other place.

The [`valueCtx`](https://github.com/golang/go/blob/02057906f7272a4787b8a0b5b7cafff8ad3024f0/src/context/context.go#L534) is derived from std.

Colstuwjx:
Totally understand.

PR #663, opened 2020-12-14
add JSONP() resp

libi:
JSONP resp should be defined by the server, and now it takes up the callback parameter


tonybase:
this is breakchange don’t merge

PR #710, opened 2021-02-18
fix http middleware

tonybase:
http middleware:
-client:
--func(ctx context.Context, req *http.Request) (*http.Response, error)
-server:
--func(ctx context.Context, req *http.Request) (http.ResponseWriter, error)


Terry-Mao:
LGTM

PR #728, opened 2021-02-27
set transport.kind to be strong type

storyicon:
There are many reasons why strong type should be used here, but only one reason that is closest to the user is described here:

The user obtains the protocol information through `transport.FromContext`, but the definition of `Transport` is as follows:
```
type Transport struct {
	Kind string
}
```

If users don’t look at the source code, they don’t know how kratos assigns a value to `Kind`. It might be：
```
grpc, gRPC, GRPC, http, HTTP
```

This makes the `Kind` field difficult to use.

Signed-off-by: storyicon <storyicon@foxmail.com>


Terry-Mao:
LGTM

storyicon:
@Terry-Mao Greate minds think alike.

PR #812, opened 2021-04-02
chore: optimize import order, refer to goimports

qloog:


tonybase:
```
import (
    stdlib

    current_project

    third_party
)
```

It should be in that order, right?

qloog:
> ```
> import (
>     stdlib
> 
>     current_project
> 
>     third_party
> )
> ```
> 
> It should be in that order, right?

mine:

```go
import (
    stdlib

    third_party

    current_project
)
```

PR #904, opened 2021-05-10
internal/http: optimize support more content_type

Cluas:
The original implementation only supports “application/sub_type*”, change it according [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1) 
```
 media-type = type "/" subtype *( OWS ";" OWS parameter )
 type       = token
 subtype    = token
```


Cluas:
好样的 go mod! 找到build失败的根源了。Lint会go get golang.org/x/lint/golint 会更新go.mod里的东西，跟go.sum不一致了。折腾好久..

PR #951, opened 2021-05-26
feature:已存在文件进行覆盖询问

Ccheers:
之前proto生成文件会自动跳过已存在文件，现在进行覆盖询问，默认跳过

### DIFF
--- a/cmd/kratos/internal/proto/server/server.go
+++ b/cmd/kratos/internal/proto/server/server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"fmt"
+	"github.com/go-kratos/kratos/cmd/kratos/v2/internal/base"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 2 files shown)


cmd/kratos/internal/proto/server/server.go
```diff
@@ -2,6 +2,7 @@ package server
 
 import (
 	"fmt"
+	"github.com/go-kratos/kratos/cmd/kratos/v2/internal/base"
```
tonybase, line 5: need to format

Ccheers: 已调整

shenqidebaozi:
I don't think it's necessary to directly cover the file, so it's better for the user to delete it manually
The generated file is a service template. Users will modify the file, and few people will repeatedly override the file

PR #999, opened 2021-06-03
optimize code generate for protoc-gen-go-http

alimy:
* this patch used for optimize code generate for protoc-gen-http then could remove some code in `encoding/json/json.go`:
  
```
// file: encoding/json/json.go

// ... other code here

func (codec) Unmarshal(data []byte, v interface{}) error {
      // ========= optimize: can remove ============ 
      // code below means to initialize an zero value that ptr point if it is nil prevant panic
	rv := reflect.ValueOf(v)
	for rv.Kind() == reflect.Ptr {
		if rv.IsNil() {
			rv.Set(reflect.New(rv.Type().Elem()))
		}
		rv = rv.Elem()
	}
    // ========= optimize: can remove ==============

	if m, ok := v.(proto.Message); ok {
		return UnmarshalOptions.Unmarshal(data, m)
	} else if m, ok := reflect.Indirect(reflect.ValueOf(v)).Interface().(proto.Message); ok {
		return UnmarshalOptions.Unmarshal(data, m)
	}
	return json.Unmarshal(data, v)
}
```

* initialize an zero value assign to an point could in generate code location when need(eg: has  body in request or need ResponseBody in response), so no need in decode method location for all sence(eg: no need ResposeBody or no need custom body filed in request).


cmd/protoc-gen-go-http/template.go
```diff
@@ -72,12 +76,12 @@ func New{{.ServiceType}}HTTPClient (client *http1.Client) {{.ServiceType}}HTTPCl
 func (c *{{$svrType}}HTTPClientImpl) {{.Name}}(ctx context.Context, in *{{.Request}}, opts ...http1.CallOption) (out *{{.Reply}}, err error) {
 	path := binding.EncodePath("{{.Method}}", "{{.Path}}", in)
 	out = &{{.Reply}}{}
-	{{if .HasBody }}
-	err = c.cc.Invoke(ctx, path, in{{.Body}}, &out{{.ResponseBody}}, http1.Method("{{.Method}}"), http1.PathPattern("{{.Path}}"))
-	{{else}} 
-	err = c.cc.Invoke(ctx, path, nil, &out{{.ResponseBody}}, http1.Method("{{.Method}}"), http1.PathPattern("{{.Path}}"))
-	{{end}}
-	return 
+    {{if ne .ResponseBody "" }}
```
alimy, line 79: * 如果需要一个自定义的responseBody filed， 以下两段代码用于初始化这个filed；
* 因为这里是按需进行初始化，所以在`encoding/json/json.go` 中的`func (codec) Unmarshal(data []byte, v interface{}) error ` 方法就不需要再做这一步逻辑了， 可以优化性能，毕竟不是每个请求都需要自定义responseBody filed，（golang反射还是会浪费一些时间的，能少一点反射调用可以优化一点性能）

PR #1018, opened 2021-06-08
modify proto-gen-go-errors version and func name

shenqidebaozi:
modify proto-gen-go-errors version and func name


tonybase (approved):
LGTM

PR #1094, opened 2021-06-22
feat(cmd): support the validate pb file generate for kratos proto cli…

luckylsx:
support the validate pb file generate for kratos command

kratos client 命令 增加 validate pb 文件生成


tonybase (approved):
LGTM

PR #1098, opened 2021-06-23
Fix cmd proto server

shenqidebaozi:
Fix the problem that the service file generated when the proto file contains google.protobuf.empty does not match the expectation。


tonybase (approved):
LGTM

PR #1102, opened 2021-06-23
fix(cmd):  fix kratos cmd

luckylsx:
fix kratos proto client cmd when the validate.proto is annotated not to generate validate pb file

修复kratos命令，validate.proto 引入，注释未使用时，不生成 validate pb文件


tonybase (approved):
LGTM

PR #1130, opened 2021-06-30
feat: add int/int32/Stringer support when get atomicValue

kagaya85:
When the variable type is `int`/`int32`, I think we can also try to convert it to the desired type, as the `int` is sometimes the default type for many conversions (such as Unmarshal), and if it is `string` we can also try to assert whether implements the `String` method.

Further more, is it reasonable to convert any other two types that is convertible (such as boolean `true` to int `1`)?

Please point out if there is anything I have not considered. Thanks in advance.


tonybase (approved):
LGTM

PR #1139, opened 2021-07-01
feat(cmd): add kratos changelog command #1133 

shenqidebaozi:
Add the changelog command in kratos cmd to obtain kratos release information and dev commits information #1133


shenqidebaozi:
The effect is shown in the figure
image
image

PR #1162, opened 2021-07-08
examples: add echo example

NSObjects:
echo example

### DIFF
--- a/examples/http/echo/main.go
+++ b/examples/http/echo/main.go
@@ -9,11 +9,11 @@ import (
 
 func main() {
 	router := echo.New()
-	router.GET("/home", func(context echo.Context) error {
-		return context.JSON(200,"Hello echo")
+	router.GET("/home", func(ctx echo.Context) error {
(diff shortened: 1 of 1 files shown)


tonybase (approved):

examples/http/echo/main.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"github.com/go-kratos/kratos/v2"
+	"github.com/go-kratos/kratos/v2/transport/http"
+	"github.com/labstack/echo/v4"
+	"log"
+)
+
+func main() {
+	router := echo.New()
+	router.GET("/home", func(context echo.Context) error {
```
tonybase, line 12: ctx

examples/http/echo/main.go
```diff
@@ -0,0 +1,28 @@
...
+		return context.JSON(200,"Hello echo")
+	})
+
+	httpSrv := http.NewServer(http.Address(":9527"))
```
tonybase, line 16: 8000

PR #1203, opened 2021-07-17
fix(transport/http): fixed the problem of getting empty node list

shenqidebaozi:
fixed the problem that the node was empty in service discovery, and update demo


transport/http/resolver.go
```diff
@@ -54,7 +54,9 @@ func newResolver(ctx context.Context, discovery registry.Discovery, target *Targ
 		watcher: watcher,
 		logger:  log.NewHelper(log.DefaultLogger),
 	}
+	done := make(chan struct{}, 1)
```
longXboy, line 57: 这里可以接受watch.Next返回的error

transport/http/resolver.go
```diff
@@ -79,8 +81,22 @@ func newResolver(ctx context.Context, discovery registry.Discovery, target *Targ
...
+				}
+			}
 		}
 	}()
+	if block {
+		select {
+		case <-done:
+			return r, nil
+		case <-ctx.Done():
+			return r, nil
```
longXboy, line 97: 如果超时了，这里返回ctx的error

transport/http/resolver.go
```diff
@@ -79,8 +81,22 @@ func newResolver(ctx context.Context, discovery registry.Discovery, target *Targ
...
+					executed = true
+					done <- struct{}{}
+				}
+			}
 		}
 	}()
+	if block {
+		select {
+		case <-done:
+			return r, nil
```
longXboy, line 95: 接受error，并返回

transport/http/resolver.go
```diff
@@ -54,7 +54,9 @@ func newResolver(ctx context.Context, discovery registry.Discovery, target *Targ
 		watcher: watcher,
 		logger:  log.NewHelper(log.DefaultLogger),
 	}
+	done := make(chan struct{}, 1)
 	go func() {
+		var executed bool
 		for {
 			services, err := watcher.Next()
 			if err != nil {
```
longXboy, line 62: 这里漏了done<-err

Issue #1219, opened 2021-07-20
config env and file source存在bug


ybdx:
#### What happened:
env and file 都是source,它们都是用过config.WithSource 传入config，如果env.NewSource()放在file.NewSource(flagconf)之后，那么config file中的变量就无法替换
			
#### What you expected to happen:
env和file平级的话，那么就不应该有先后次序

#### Environment:
- Kratos version (use `kratos -v`):
- Go version (use `go version`):
- OS (e.g: `cat /etc/os-release`):
- Others:

PR #1220, opened 2021-07-20
fix(config): load env and file source order bug

EndlessIdea:
When you want to use env vars to replace the placeholders in config file, you must put env source first to init the config. It's not very friendly to users.
Changed the error message to lowercase to comply with Go's error standard.

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -81,7 +81,7 @@ func defaultResolver(input map[string]interface{}) error {
 		} else if len(args) > 1 { // default value
 			return args[1]
 		}
-		return ""
+		return args[0]
 	}
 
 	var resolve func(map[string]interface{}) error
--- a/config/options.go
+++ b/config/options.go
@@ -81,7 +81,7 @@ func defaultResolver(input map[string]interface{}) error {
 		} else if len(args) > 1 { // default value
 			return args[1]
 		}
-		return args[0]
+		return ""
 	}
 
 	var resolve func(map[string]interface{}) error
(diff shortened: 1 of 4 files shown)


shenqidebaozi:
#1219

ymh199478 (approved):
LGTM

ymh199478:
等下，如果这样调整，watch 得到的配置是不能够替换的

kagaya85:
IMHO, it's reasonable using `resolver` after all config loaded, but it also causes some value can not update when watching value change to environment variable placeholder in runtime, it may acceptable and requires that user should use placeholder when service start.

otherwise, we may also add `resolver` when each keyvalue merge in, for this reason, not found value should use origin value (such as `${key}`) instead of empty string `""` :)
https://github.com/go-kratos/kratos/blob/55599231939c89ba810950361d0a2b570d2865d8/config/options.go#L75-L85

EndlessIdea:
Thanks for pointing this out! I add the resolve after it merges the next config and return the placeholder if its value is not found.

config/options.go
```diff
@@ -81,7 +81,7 @@ func defaultResolver(input map[string]interface{}) error {
 		} else if len(args) > 1 { // default value
 			return args[1]
 		}
-		return ""
+		return args[0]
```
kagaya85, line 84: If `resolver` is added to each watch loop, it's OK to return empty `""` for not found value, because env vars are not support to be added by `watch` now, and new added palceholder will not be resolved again.

EndlessIdea: got it! I misunderstand your last comment haha

ymh199478:
I think the `watch` behavior should be protected by adding a unit test.

ymh199478 (approved):
LGTM

PR #1306, opened 2021-08-06
feat(cmd/pgh): add processing when generating HTTP code

shenqidebaozi:
For interfaces that require body, such as POST、PUT、PATCH, if no body is declared, the warning log is printed


shenqidebaozi:
In addition, body binding should not be allowed in delete requests

Windfarer (approved):
lgtm

longXboy (approved):

PR #1321, opened 2021-08-12
feat(cmd/new): replace greeter with the project name

shenqidebaozi:
replace greeter with the project name


shenqidebaozi:
@Windfarer Whether the name in the template needs to be configured through parameters？Such as “Greeter”

PR #1355, opened 2021-08-16
fix: correct client code in the example of TLS connection

iamharvey:

A public certificate should be loaded instead of using the server certificate. It could be a root CA cert or an intermediate CA cert. The example may mislead those who are not familiar with TLS practice.

#### Which issue(s) this PR fixes:

fixes #


longXboy:
Please replace <> with the content of the CA certificate actually generated, which is the last paragraph in server.crt

examples/tls/cert/ca.crt
```diff
@@ -0,0 +1,3 @@
+-----BEGIN CERTIFICATE-----
+<pem content>
```
longXboy, line 2: Please replace <pem content> with the content of the CA certificate actually generated, which is the last paragraph in server.crt

iamharvey:
@longXboy fixed

examples/tls/cert/ca.crt
iamharvey: fixed

longXboy (approved):

PR #1359, opened 2021-08-17
feat!: add Watchable interface

letian0805:
fix: typo
feat: add Watchable

Config source has a method "Watch", but some sources cannot be watched and can only be loaded.So it is necessary to add a Watchable interface.
#### Which issue(s) this PR fixes:

fixes #


ymh199478:
For some sources, watch can do nothing.  in other words watch has no behavior.

Is there any other reason do you need to distinguish between them (loader and source)?

kagaya85:
Separate them seems to provide no visible benefit but rather more complexity, I think it makes the user confused about when to use `NewSource` or `NewLoader`. For the same function (load config), we want to provide as an simple interface as possible and should not make users waste time on how to choose them.

letian0805:
> For some sources, watch does nothing. in other words watch has no behavior.
> 
> Is there any other reason do you need to distinguish between them (loader and source)?

User is developer. They may load configure from other sources, those don't need watch. If use Source, they need implement empty Watch interface. This looks even more confusing. Analogous to the io package in the Go standard library, which provides both io.Closer and io.ReadCloser, this is the reason.

letian0805:
> Separate them seems to provide no visible benefit but rather more complexity, I think it makes the user confused about when to use `NewSource` or `NewLoader`. For the same function (load config), we want to provide as an simple interface as possible and should not make users waste time on how to choose them.

Analogous to the io package in the Go standard library, which provides both io.Closer and io.ReadCloser, this is the reason.

config/config.go
```diff
@@ -107,13 +107,15 @@ func (c *config) Load() error {
 			c.log.Errorf("failed to merge config source: %v", err)
 			return err
 		}
-		w, err := src.Watch()
-		if err != nil {
-			c.log.Errorf("failed to watch config source: %v", err)
-			return err
+		if s, ok := src.(Source); ok {
```
kagaya85, line 110: Perhaps we can take a discussion of the specific interface, for a config source, `Load()` method is necessary to implement however `Watch()` is not, so the `Watch()` method can be independent from `Source` interface, such as an interface named `Watchable`, and keep the `NewSource` function to maintain the existing usage unchanged. 

```go
type Source interface {
    Load() ([]*KeyValue, error)
}

type Watchable interface {
    Watch() (Watcher, error)
}
```

we can assert here whether config source implements the `Watchable` interface, we'd better not add a new entry function like `NewLoader`, what do you think about it?

letian0805: Yes, you are right. This is better.

config/env/env.go
```diff
@@ -7,30 +7,38 @@ import (
...
+var _ config.Source = (*envSource)(nil)
+
 type env struct {
-	prefixs []string
+	prefixes []string
+}
+
+type envSource struct {
+	*env
 }
```
kagaya85, line 18: Is `envSource` necessary?

config/env/env_test.go
```diff
@@ -367,7 +367,7 @@ func Test_env_load(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			e := &env{
-				prefixs: tt.fields.prefixs,
+				prefixes: tt.fields.prefixs,
```
kagaya85, line 370: also use `tt.fields.prefixes`?

config/options.go
```diff
@@ -28,7 +28,7 @@ type options struct {
 // WithSource with config source.
 func WithSource(s ...Source) Option {
 	return func(o *options) {
-		o.sources = s
+		o.sources = append(o.sources, s...)
```
kagaya85, line 31: Maybe the original `WithSource` is enough？

kagaya85:
PTAL @tonybase

config/env/env.go
letian0805: Removed

config/env/env_test.go
letian0805: Updated

config/options.go
letian0805: User maybe use WithSource many times with different Source

kagaya85: `WithSource` can accept multiple parameters, what are the scenarios that require the use of multiple `WithSource`?

letian0805: User don't know WithSource can only use once with all sources, they maybe use twice and more with different sources. The framework should be comfortable for users to use.

kagaya85:
Thanks for contribution, basically LGTM.

ymh199478 (approved):
LGTM，there is a potential **BC**, see if it needs to be marked

letian0805:
@tonybase  Please review this PR.

tonybase:
There is a BC and splitting the interface is not recommended.

PR #1391, opened 2021-08-30
fix: ci lint error

kagaya85:

The new linters casused many static check errors, we need to fix them all.


shenqidebaozi (approved):
LGTM

Issue #1430, opened 2021-09-06
[Help Wanted] fix lint check failures


kagaya85:
After PR #1401 merged, many of our sub modules are failing lint, we should fix all of them.

To fix these, take a look at [hack/.lintcheck_failures](https://github.com/go-kratos/kratos/blob/main/hack/.lintcheck_failures) for a list of currently lint failed modules.

You can check a module by removing it from the `.lintcheck_failures` list and running `make fix` to auto fix some lint errors.

If the module is no longer failing, please file a PR including the fixes and removing it from the list.

Before filing your PR, you should run `make lint` to check your code is already fixed and please keeping PRs scoped down in size for review, fix a package or set of closely related package.

**NOTICE:** 
* Do not use `fixes #...` in your PR, that will close this PR. Instead put something like `part of #1430`
* To attract more contributors in our community, I recommend letting new contributors do this as much as possible

Modules should be fixed:
- [x] ./examples https://github.com/go-kratos/kratos/pull/1454
- [x] ./cmd/protoc-gen-go-errors https://github.com/go-kratos/kratos/pull/1435
- [x] ./cmd/protoc-gen-go-http https://github.com/go-kratos/kratos/pull/1440
- [x] ./cmd/kratos https://github.com/go-kratos/kratos/pull/1442
- [x] ./contrib/metrics/datadog https://github.com/go-kratos/kratos/pull/1517
- [x] ./contrib/config/nacos https://github.com/go-kratos/kratos/pull/1522
- [x] ./contrib/config/apollo https://github.com/go-kratos/kratos/pull/1523
- [x] ./contrib/config/kubernetes https://github.com/go-kratos/kratos/pull/1523
- [x] ./contrib/registry/nacos https://github.com/go-kratos/kratos/pull/1599
- [x] ./contrib/registry/consul https://github.com/go-kratos/kratos/pull/1530
- [x] ./contrib/registry/kubernetes https://github.com/go-kratos/kratos/pull/1599
- [x] ./contrib/registry/zookeeper https://github.com/go-kratos/kratos/pull/1599
- [x] ./contrib/registry/etcd https://github.com/go-kratos/kratos/pull/1441
- [x] ./contrib/log/fluent https://github.com/go-kratos/kratos/pull/1432

ShawnAli97:
Hello I am new to this community. Can I make a attempt at fixing a module?

kagaya85:
> Hello I am new to this community. Can I make a attempt at fixing a module? 

[...]

PR #1441, opened 2021-09-07
fix: Fix lint of files in contrib/registry/etcd

Casper-Mars:
part of #1430
PTAL @tonybase. I add `error` return value to func `newWatcher`. Should I define a new error  instead of sentinel error?


tonybase (approved):

kagaya85:
@Casper-Mars Too much thanks to your contributions, I am very glad to see these modules has been fixed and I have some ideas: 

Could we leave the rest of opportunity to more newcomers？so that we can attract more participants into our community.

Thanks again

Casper-Mars:
@kagaya85 Of course. It is my pleasure.

PR #1512, opened 2021-09-24
encoding/form: support repeated filed map from path etc.

Cluas:

https://github.com/go-kratos/kratos/issues/1193

#### Other special notes for reviewer:

Now there is no compulsion to use the same legal separator, which means that the user can use the format `a b,c|d\te`, which will also be legally parsed as `[]string{"a", "b", "c", "d", "e"}`, I'm not sure if this is good practice, maybe we can restrict it to a single legal separator.


Cluas:
Considering the complexity, it is recommended to first pass the array via `parameter` instead of `path`, something like `id=1&id=2`, to temporarily turn off this PR.

PR #1528, opened 2021-09-30
fix: kubernetes lint

NSObjects:
part of #1430


kagaya85:
kubernetes config lint error is already fixed by https://github.com/go-kratos/kratos/pull/1523, thanks for contribution

PR #1534, opened 2021-10-03
fix: 'tokenInfo' might have 'nil' or bad value

ffy:

'tokenInfo' might have 'nil' or other unexpected value as its corresponding error variable might be not 'nil'


Casper-Mars:
Hi~I believe the origin `err` info should return. Try to define an err that contain the origin err message. Maybe something like this:`errors.Unauthorized(reason, err.Error())`

ffy:
> Hi~I believe the origin `err` info should return. Try to define an err that contain the origin err message. Maybe something like this:`errors.Unauthorized(reason, err.Error())`

There u go! The code has been updated.

tonybase (approved):

PR #1544, opened 2021-10-13
fix(protoc-gen-go-errors): fix the camel case bug of name

storyicon:
Signed-off-by: storyicon <yuanchao@bilibili.com>


storyicon:
cc @tonybase

PR #1553, opened 2021-10-15
fix(cmd/proto-gen-http): fix the problem that the field declaration does not exist causing panic

shenqidebaozi:
…oes not exist causing panic

fixes #


shenqidebaozi:
合并注意修改一下 PR 信息
 fix(cmd/proto-gen-http): fix the problem that the field declaration does not exist causing panic

PR #1620, opened 2021-11-10
fix:add consul all health check

haiyux:

fixes #

### DIFF
--- a/contrib/registry/consul/client.go
+++ b/contrib/registry/consul/client.go
@@ -98,8 +98,8 @@ func (d *Client) Register(ctx context.Context, svc *registry.ServiceInstance, en
 		for _, address := range checkAddresses {
 			asr.Checks = append(asr.Checks, &api.AgentServiceCheck{
 				TCP:                            address,
-				Interval:                       "30s",
-				DeregisterCriticalServiceAfter: "90s",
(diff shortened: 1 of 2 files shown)


contrib/registry/consul/client.go
```diff
@@ -71,13 +71,15 @@ func (d *Client) Register(ctx context.Context, svc *registry.ServiceInstance, en
 	addresses := make(map[string]api.ServiceAddress)
 	var addr string
 	var port uint64
```
tonybase, line 73: `Addr, port` can actually be removed, and the default asr.Checks = {}.

haiyux: Do not use health check?

contrib/registry/consul/client.go
```diff
@@ -86,8 +86,6 @@ func (d *Client) Register(ctx context.Context, svc *registry.ServiceInstance, en
 		Meta:            svc.Metadata,
 		Tags:            []string{fmt.Sprintf("version=%s", svc.Version)},
 		TaggedAddresses: addresses,
-		Address:         addr,
```
longXboy, line 89: why Address  was deleted in registration?

tonybase (approved):

PR #1629, opened 2021-11-12
fix(cmd/protoc-gen-go-http): follow the http rule to use query strings

bakjos:
This PR fixes the implementation of the HTTP generator to follow the rule specified in the [http.proto](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L215) file, of assigning the query parameters to the messages, by binding the query string after the body is bound


tonybase (approved):
LGTM

shenqidebaozi (approved):

PR #1660, opened 2021-12-02
fix(grpc): Block the RCP unil a new picker is available

bakjos:
This PR fixes a bug where it throws a no_available_node error when there are no nodes to select from. The pick function now will return a ErrNoSubConnAvailable to block the RPC until a new picker is available via UpdateState().

fixes #


tonybase (approved):
LGTM

haiyux (approved):
LGTM

PR #1704, opened 2021-12-21
fix(protoc-gen-go-http): Update http rule when path mappings

bakjos:
When generating the HTTP code including path mappings, like 

```proto
   service Messaging {
       rpc GetMessage(GetMessageRequest) returns (Message) {
         option (google.api.http) = {
             get: "/v1/{name=messages/*}"
         };
       }
  }
```

It was not detecting `name` as a variable, so this PR switches to use regex instead, it also creates an additional rule for these cases to support what's described in the HTTP rule https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L85

fixes #


longXboy:
in this pr, if the request uri is "/v1/messages/123456",  name's value still is "123456" not "messages/123456".
is that expected?

bakjos:
@longXboy Thanks for the quick response, it is expected in the way I made the PR, but not sure if it should work in that way, I think it should match what you get on the GRPC method so it should be message/12346 instead of 123456, I will make the change, although it implies modifying how the variables are capture in the http module

bakjos:
@longXboy I think I found a way to match the route without the need of having 2 routes and following the guidelines, let me know what you think

longXboy:
Sorry for the late reply, now the implementation is more elegant and standard, LGTM!
@bakjos

bakjos:
@longXboy wasn't sure where to add tests, can you point me out to where I can add them?

longXboy:
> @longXboy wasn't sure where to add tests, can you point me out to where I can add them?. I wanted to add one, but couldn't find any for protocol-gen-go-http

just add some test case for Function `replacePath ` and `buildPathVars`

bakjos:
> > @longXboy wasn't sure where to add tests, can you point me out to where I can add them?. I wanted to add one, but couldn't find any for protocol-gen-go-http
> 
> @bakjos just add some test cases for Function `replacePath ` and `buildPathVars`

@longXboy  Done, thanks for the guidence

longXboy:
@bakjos  github ci did not pass,Maybe you forgot to execute
 `go get github.com/stretchr/testify/assert@v1.7.0` or `go mod tidy`

bakjos:
> @bakjos github ci did not pass,maybe you forgot to execute `go get github.com/stretchr/testify/assert@v1.7.0` or `go mod tidy`

@longXboy  Yup, forgot to include the go.mod file. Thanks

longXboy (approved):

haiyux (approved):

longXboy:
fix  #1653

PR #1717, opened 2021-12-27
feat: add pool Transport

haiyux:

fixes #


Ehco1996:
just a question, is there any bechmark or test report can show use pool transport have better performance?

haiyux:
> just a question, is there any bechmark or test report can show use pool transport have better performance?

Using pool to avoid duplicate object creation and reduce GC pressure, it will certainly improve performance.
Many open source frameworks do this, such as gin's context.

tonybase (approved):

PR #1724, opened 2021-12-28
fix(encoding/form): FieldMask are converted to/from lower-camel naming conventions.

tonybase:
// The JSON representation for a FieldMask is a JSON string where paths are
// separated by a comma. Fields name in each path are converted to/from
// lower-camel naming conventions. Encoding should fail if the path name would
// end up differently after a round-trip.

references: https://github.com/protocolbuffers/protobuf-go/blob/master/encoding/protojson/well_known_types.go#L842


shenqidebaozi (approved):
LGTM

longXboy (approved):

PR #1497, opened 2021-09-20
Feat/refactor endpoint

longXboy:
fix #1494
fix #1413
1.add listener options
2.remove Endpoint() side effect

PR #1470, opened 2021-09-13
feat(transport): add listener option

letian0805:
#### Description (what this PR does / why we need it):

net.Listener is an interface, which allows users to implement it themselves. For example, users can wrap the net.Listener of the standard library to do something when receiving a connection, such as outputting logs. The framework should provide such extensibility as the standard library.

#### Which issue(s) this PR fixes (resolves / be part of):

fixes #1413

PR #1729, opened 2021-12-29
feat(transport):add grpc and http with listen

haiyux:

fixes #

### DIFF
--- a/transport/grpc/server.go
+++ b/transport/grpc/server.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"net"
 	"net/url"
-	"sync"
 	"time"
 
 	"github.com/go-kratos/kratos/v2/internal/endpoint"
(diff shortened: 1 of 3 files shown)


transport/grpc/server.go
```diff
@@ -150,21 +156,6 @@ func NewServer(opts ...ServerOption) *Server {
 // examples:
 //   grpc://127.0.0.1:9000?isSecure=false
 func (s *Server) Endpoint() (*url.URL, error) {
-	s.once.Do(func() {
-		lis, err := net.Listen(s.network, s.address)
```
Casper-Mars, line 154: would be it satisfy if just replace this line by s.lis(if s.lis is not nil)?

haiyux: Yes，I want to get rid of the side effects of endpoint。

letian0805:
Hey! This is the third PR about the side effects of Listener and Endpoint, right?
The first one is https://github.com/go-kratos/kratos/pull/1470

haiyux:
> Hey! This is the third PR about the side effects of Listener and Endpoint, right? The first one is #1470

Your PR does not eliminate the side effects of endpoint, it still listen in the endpoint function.

letian0805:
> > Hey! This is the third PR about the side effects of Listener and Endpoint, right? The first one is #1470
> 
> Your PR does not eliminate the side effects of endpoint, it still listen in the endpoint function.

Yes i know, so longxboy created another one to fix side effects. https://github.com/go-kratos/kratos/pull/1497

haiyux:
feat #1714

tonybase (approved):

PR #1738, opened 2021-12-31
feat(registry/consul): add WithHeartbeat option

letian0805:

Add WithHeartbeat option, so user can choose healthcheck or heartbeat, or both.

fixes #


letian0805 (commented):
@longXboy @haiyux

tonybase (approved):

PR #1761, opened 2022-01-11
feat(log): add global logger appliance, as process level default logger

realityone:

Add global logger appliance, make logging in independence function easily.

fixes #

### DIFF
--- a/log/global.go
+++ b/log/global.go
@@ -1,0 +1,126 @@
+package log
+
+import (
+	"fmt"
+	"os"
+	"sync"
+)
+
+// globalLogger is designed as a global logger in current process.
+var (
+	global = &loggerAppliance{}
+)
+
+type loggerAppliance struct {
+	lock   sync.Mutex
+	logger Logger
+	helper *Helper
+}
+
+func init() {
+	global.SetLogger(NewStdLogger(os.Stderr))
+}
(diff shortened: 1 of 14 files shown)


shenqidebaozi (approved):
LGTM

log/global.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+type loggerAppliance struct {
+	lock   sync.Mutex
+	logger Logger
+	helper *Helper
+}
+
+func init() {
+	global.SetLogger(NewStdLogger(os.Stderr))
```
tonybase, line 21: Use `log.DefaultLogger` or mark it `//Deprecated` and use GetLogger instead

log/global.go
```diff
@@ -0,0 +1,126 @@
...
+	a.helper = NewHelper(a.logger)
+}
+
+func (a *loggerAppliance) GetLogger() Logger {
+	return a.logger
+}
+
+// SetGlobalLogger should be called before any other log call.
+// And it is not THREAD SAFE.
+func SetGlobalLogger(logger Logger) {
```
tonybase, line 37: Is it better to use `log.SetLogger(logger)`?

tonybase (approved):

PR #1822, opened 2022-02-16
feat: allow specify namespace for k8s registry

ctrysbita:
Currently the k8s registry uses default option for informer, which requires the global permission to list and watch pods for all namespaces. This PR adds a custom constructor to specifiy custom namespace without breaking the original one.

### DIFF
--- a/contrib/registry/kubernetes/registry.go
+++ b/contrib/registry/kubernetes/registry.go
@@ -89,9 +89,18 @@ type Registry struct {
 	stopCh chan struct{}
 }
 
-// NewRegistry is used to initialize the Registry
+// NewRegistry initialize the k8s registry with global namespace.
 func NewRegistry(clientSet *kubernetes.Clientset) *Registry {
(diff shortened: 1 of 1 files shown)


contrib/registry/kubernetes/registry.go
```diff
@@ -89,9 +89,18 @@ type Registry struct {
...
 
-// NewRegistry is used to initialize the Registry
+// NewRegistry initialize the k8s registry with global namespace.
 func NewRegistry(clientSet *kubernetes.Clientset) *Registry {
-	informerFactory := informers.NewSharedInformerFactory(clientSet, time.Minute*10)
+	return NewRegistryWithNamespace(clientSet, corev1.NamespaceAll)
+}
+
+// NewRegistry initialize the k8s registry with specific namespace.
+func NewRegistryWithNamespace(clientSet *kubernetes.Clientset, namespace string) *Registry {
```
tonybase, line 98: func NewRegistry(client, ...Option)

Should be provide `Option` configuration

PR #1828, opened 2022-02-20
Feature(WIP): implement a Kafka server

EndlessIdea:
feat[server]: Implement a Kafka server 

This PR is to add a Kafka server which acts as a job worker to consume messages received from Kafka. It uses "github.com/Shopify/sarama" as the Golang Kafka library.

A demo here https://github.com/EndlessIdea/kratos-kafka-server-demo
Please see the usage in internal/service, internal/server and cmd/server.


contrib/server/kafka/server.go
```diff
@@ -0,0 +1,113 @@
...
+	if len(server.handlers) == 0 {
+		return nil, fmt.Errorf("no handlers")
+	}
+
+	for _, consumer := range server.consumers {
+		for _, handler := range server.handlers {
+			if consumer.Topic() != handler.Topic() {
+				continue
+			}
+			consumer.RegisterHandler(handler)
```
Casper-Mars, line 82: Why register handler at this point? I think someone can register the handler before passing the consumer.

EndlessIdea: Yes, this way can't prevent the user register the handler themselves. The reason I write this way is I want to use wire to generate the NewServer code. I have a demo code here
```
// NewKafkaServer creates a new Kafka Server
func NewKafkaServer(config *conf.Server, handlers []kafka.Handler) (*kafka.Server, error) {
	// init consumers
	var consumers []kafka.Consumer
	brokers := config.Kafka.Brokers
	for _, cConf := range config.Kafka.Consumers {
		srvConsumer, err := consumer.NewGroupConsumer(brokers, cConf.Topics, cConf.Group)
		if err != nil {
			return nil, errors.WithMessage(err, "init consumer error")
		}
		consumers = append(consumers, srvConsumer)
	}

	return kafka.NewServer([]kafka.ServerOption{
		kafka.Consumers(consumers),
		kafka.Handlers(handlers),
	}...)
}

```

Casper-Mars: Oh~I see.But I think wire does not allow multiple providers for one type as mentioned on this [doc](https://github.com/google/wire/blob/main/docs/faq.md#what-if-my-dependency-graph-has-two-dependencies-of-the-same-type)

EndlessIdea: We could return a slice of handlers or consumers, I have a kafka-server-demo here https://github.com/EndlessIdea/kratos-kafka-server-demo.

contrib/server/kafka/server.go
```diff
@@ -0,0 +1,119 @@
...
+func ErrorHandlers(errorHandlers []Handler) ServerOption {
+	return func(server *Server) {
+		for _, handler := range errorHandlers {
+			server.errorHandlers[handler.Topic()] = handler
+		}
+	}
+}
+
+// NewServer creates a Kafka server by options.
+func NewServer(consumers []Consumer, handlers []Handler, opts ...ServerOption) (transport.Server, error) {
```
Casper-Mars, line 57: In my opinion, handlers and consumers are closely related. User should register the handler on their own.Relating consumer and handler by topic is not precise. Sometime two consumers consume same topic from different kafka.

EndlessIdea: Sounds reasonable! If then, the NewConsumer should be changed as 
`func NewConsumer(brokers []string, topic, group string, handler kafka.Handler, opts ...ConfigOption) (*Consumer, error) `.
And we won't be able to initialize the Server via wire because it doesn't know which handler should be used to initialize which consumer.

Casper-Mars: I don't think whether or not we use Wire has anything to do with the framework. User still can inject dependence by Wire. For example, they can inject consumer instead of handler. It is just the matter of how to use Wire.

Casper-Mars: https://cloud.google.com/solutions/event-driven-architecture-pubsub

PR #1848, opened 2022-02-28
refactor(middleware/logging): use Error method instead of fmt.Sprintf

sashamelentyev:

fixes #


haiyux (approved):

haiyux (changes requested):
If use `github.com/pkg/errors` to wrap error , `fmt.Sprintf("%+v", err)` can return stack information.

PR #1863, opened 2022-03-05
feat: first version

tx7do:

fixes #


Casper-Mars:
Hi~ Thanks for your contributions. I think it's better to push to beer-shop or blog.

kagaya85:
hi @tx7do , we already move the examples code to a new repo, see [here](https://github.com/go-kratos/examples), other than that, I think we need to keep the examples as minimally usable as possible and avoid some license issues, It does not reasonable to directly quote someone's code with license. 

Welcome to open another PR in our new example repo, I will close this, thanks

PR #1899, opened 2022-03-29
feat(log): log load config only when Debug level

Cluas:

Normally we don't need to know the config load key info, only we need to be aware of it when debugging.
So change Info level to Debug

fixes #

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -104,7 +104,7 @@ func (c *config) Load() error {
 			return err
 		}
 		for _, v := range kvs {
-			c.log.Infof("config loaded: %s format: %s", v.Key, v.Format)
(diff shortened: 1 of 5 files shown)


tonybase (dismissed):

Cluas:
It looks like there is a problem with some tests, can you help look at them? @huyuanxin

huyuanxin:
> It looks like there is a problem with some tests, can you help look at them? @huyuanxin

ok, I will fix docker image

Cluas:
> > It looks like there is a problem with some tests, can you help look at them? @huyuanxin
> 
> ok, I will fix docker image
> 
> update: Now, docker image is fixed, please try run test actions again @Cluas

Thanks

shenqidebaozi (approved):

haiyux (approved):

PR #1900, opened 2022-03-29
add contrib registry zeroetcd for go-zero

fengjiannan2010:
add contrib registry zeroetcd for go-zero

fixes #


haiyux:
https://github.com/haiyux/kratos-zerorule-etcd

Issue #1897, opened 2022-03-23
middleware log filter会丢失初始logger 的 prefix设置，导致一些值获取不到


ducknightii:
#### What happened:

middleware tracing + logging filter 一起使用时，日志中打印的trance_id 丢失

#### What you expected to happen:

日志中打印的trance_id存在

#### How to reproduce it (as minimally and precisely as possible):

main.Run:
```
logger := log.With(log.NewStdLogger(os.Stdout),
		"ts", log.DefaultTimestamp,
		"caller", log.Caller(4),
		"service.id", internal.ID,
		"service.name", internal.Name,
		"service.version", internal.Version,
		"trace_id", tracing.TraceID(),
		"span_id", tracing.SpanID(),
	)
```
server.NewHTTPServer:
```
// NewHTTPServer new a HTTP server.
func NewHTTPServer(c *conf.Server,  svc *service.Service, logger log.Logger) *http.Server {
	var opts = []http.ServerOption{
		http.Middleware(
			recovery.Recovery(),
			metrics.Server(
				metrics.WithSeconds(prom.NewHistogram(_metricSeconds)),
				metrics.WithRequests(prom.NewCounter(_metricRequests)),
			),
			tracing.Server(),
			logging.Server(log.NewFilter(logger,
				log.FilterFunc(func(level log.Level, keyvals ...interface{}) bool {
					for i := 0; i < len(keyvals); i++ {
						if keyvals[i] == "args" {
							keyvals[i+1] = "****"
						}
					}
					return false
				}),
			)),
		),
	}
	if c.Http.Network != "" {
		opts = append(opts, http.Network(c.Http.Network))
	}
	if c.Http.Addr != "" {
		opts = append(opts, http.Address(c.Http.Addr))
	}
	if c.Http.Timeout != nil {
		opts = append(opts, http.Timeout(c.Http.Timeout.AsDuration()))
	}
	srv := http.NewServer(opts...)
	v1.RegisterSearchHTTPServer(srv, svc)
	return srv
}

```

#### Environment:
- Kratos version (use `kratos -v`): v2.2.0
- Go version (use `go version`): 1.17.6
- OS (e.g: `cat /etc/os-release`): ubuntu 18.04
- Others:

shenqidebaozi:
@Cluas

PR #1901, opened 2022-03-29
fix(log): FilterFunc keyvals lost logger prefix

Cluas:

fixes # https://github.com/go-kratos/kratos/issues/1897

### DIFF
--- a/log/filter.go
+++ b/log/filter.go
@@ -64,7 +64,16 @@ func (f *Filter) Log(level Level, keyvals ...interface{}) error {
 	if level < f.level {
 		return nil
 	}
-	if f.filter != nil && f.filter(level, keyvals...) {
+	var fkv []interface{}
+	// is use logger and with prefix,
+	// carry them to filer keyvals also.
+	if l, ok := f.logger.(*logger); ok {
+		if len(l.prefix) > 0 {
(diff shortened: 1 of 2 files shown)


kagaya85:
we may need to add some unit tests?

log/filter.go
```diff
@@ -64,7 +64,19 @@ func (f *Filter) Log(level Level, keyvals ...interface{}) error {
 	if level < f.level {
 		return nil
 	}
-	if f.filter != nil && f.filter(level, keyvals...) {
+	var fkv []interface{}
+	// is use logger and with prefix,
+	// carry them to filer keyvals also.
```
kagaya85, line 69: LGTM only with some typos, it may occur unavoidable memory copies and I can't see any better solution either :)

```suggestion
	// fkv is used to provide a logger slice to contains both prefix and keyvals for filter
```

Cluas: done

Cluas:
> we may need to add some unit tests?

This test is quite difficult to write, 😂, anyway, I added some tests. @kagaya85

kagaya85 (approved):

shenqidebaozi (approved):

PR #1939, opened 2022-04-19
feat(registry): support multi registrar

### LINKED ISSUE #1940: [Feature]Support multi registrar
### What problem is the feature used to solve?

Support multi registrars

letian0805:

Sometimes we need to register services to multiple registries, such as: registering to multiple consul, multiple aws elb, etc.

fixes #1940


tonybase:
You should implement multiple registries for compatible internal migration in the Discovery SDK

letian0805:
> You should implement multiple registries for compatible internal migration in the Discovery SDK

Why change the SDK? Now this RegistrarGroup implementation is for compatibility and does not need to change the SDK. Users can create multiple registries and then use RegistrarGroup to combine them into one and put them in the App parameters.

letian0805:
This is not to register multiple instances, but to register instances to multiple registries.

PR #1945, opened 2022-04-24
fix(transport/http): responseEncoder should not write any data when it need to write nil

### LINKED ISSUE #1946: ResponseEncode should not write any data when it encode nil
#### What happened:
When I don't need to return data to client
1. When using the following code, the DefaultResponseEncode will report an error(http: request method or response status code does not allow body)
```go
func Handle(ctx http.Context) error {
    return ctx.Result(204, nil)
}
```
2. but, when using the following code, will return `null` string to client
```go
func Handle(ctx http.Context) error {
    return ctx.Result(200, nil)
}
```

when call ctx.Result(xxx, nil), server don't return any data to client 

#### Environment:
- Kratos version (use `kratos -v`): `kratos version v2.2.1`
- Go version (use `go version`): 
- OS (e.g: `cat /etc/os-release`):
- Others:

Donglong:

It should not write any data when we don't need to return data to the client

fixes #1946

### DIFF
--- a/transport/http/codec.go
+++ b/transport/http/codec.go
@@ -42,14 +42,14 @@ func DefaultRequestDecoder(r *http.Request, v interface{}) error {
 
 // DefaultResponseEncoder encodes the object to the HTTP response.
 func DefaultResponseEncoder(w http.ResponseWriter, r *http.Request, v interface{}) error {
-	var data []byte
-	var err error
+	if v == nil {
+		_, err := w.Write(nil)
+		return err
+	}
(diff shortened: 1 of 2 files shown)


shenqidebaozi:
You should submit an issue and explain in the issue what problems you have encountered and why you want to solve them in this way, and then associate the PR with it to facilitate other users to understand what the PR has done

Donglong:
@shenqidebaozi I done, is OK?

tonybase (approved):

PR #1964, opened 2022-04-29
feat(registry): contrib/registry/zookeeper add digest acl support

youzhixiaomutou:

The current registrar with zookeeper doesn't have any method for auth.
I think add a simple digest acl support is necessary.

### DIFF
--- a/contrib/registry/zookeeper/register.go
+++ b/contrib/registry/zookeeper/register.go
@@ -44,16 +44,10 @@ func WithTimeout(timeout time.Duration) Option {
 	return func(o *options) { o.timeout = timeout }
 }
 
-// WithUser with registry user.
-func WithUser(user string) Option {
+// WithDigestACL with registry password.
+func WithDigestACL(user string, password string) Option {
 	return func(o *options) {
 		o.user = user
-	}
-}
-
-// WithPassword with registry password.
-func WithPassword(password string) Option {
-	return func(o *options) {
 		o.password = password
 	}
 }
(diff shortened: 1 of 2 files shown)


contrib/registry/zookeeper/register.go
```diff
@@ -42,6 +44,20 @@ func WithTimeout(timeout time.Duration) Option {
...
+	}
+}
+
+// WithPassword with registry password.
+func WithPassword(password string) Option {
+	return func(o *options) {
+		o.password = password
+	}
+}
+
```
Cluas, line 60: Do user passwords always come together, so why don't we put them together？

youzhixiaomutou: like this?
```go
// WithDigestACL with registry password.
func WithDigestACL(user string, password string) Option {
	return func(o *options) {
		o.user = user
		o.password = password
	}
}
```

contrib/registry/zookeeper/register.go
```diff
@@ -44,16 +44,10 @@ func WithTimeout(timeout time.Duration) Option {
 	return func(o *options) { o.timeout = timeout }
 }
 
-// WithUser with registry user.
-func WithUser(user string) Option {
+// WithDigestACL with registry password.
+func WithDigestACL(user string, password string) Option {
 	return func(o *options) {
 		o.user = user
```
Cluas, line 50: How about merging these two directly into digestACL？

Cluas: WithDigestACL build user, password to  digestACL

Cluas (approved):
LGTM

Cluas:
Thanks for the contribution!

PR #1978, opened 2022-05-13
feat: etcd config supports multiple keys

i6u:

etcd config supports multiple keys

like this:
```bash
# with prefix
/kratos/test/db.yaml
/kratos/test/cache.yaml
/kratos/test/app
```

### Implementation mode

compatible with existing versions

example:
```go
func WithPath(paths ...string) Option {
	return func(o *options) {
		o.paths = paths
	}
}
````

### Usage demonstration

example:
```go
source, err := etcd.New(client, WithPrefix(true), WithPath("/kratos/test/db.yaml",
                                                   "/kratos/test/cache.yaml",
                                                   "/kratos/test/app"))
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -31,7 +31,7 @@ type Observer func(string, Value)
 // Config is a config interface.
 type Config interface {
 	Load() error
-	Scan(v interface{}) error
+	Scan(v ...interface{}) error
 	Value(key string) Value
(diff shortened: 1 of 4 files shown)


tonybase:
Multiple configurations should be consolidated into one and unmarshal to config struct.

PR #1982, opened 2022-05-16
fix: no need using WaitGroup

reason-tech:

fixes #


app.go
```diff
@@ -94,13 +94,12 @@ func (a *App) Run() error {
 			defer cancel()
 			return srv.Stop(stopCtx)
 		})
-		wg.Add(1)
+
 		eg.Go(func() error {
-			wg.Done()
 			return srv.Start(NewContext(a.opts.ctx, a))
```
tonybase, line 99: How to register after the startup is complete

PR #1992, opened 2022-05-17
Feature/proto gen go http adding register with prefix path

toeydevelopment:
Description: for more dynamic register url path i added a new template function call Register{{.ServiceType}}HTTPServerWithPrefixPath(s *http.Server, srv {{.ServiceType}}HTTPServer, prefixPath string)


toeydevelopment:
test-coverage still failed even on main branch

toeydevelopment:
@tonybase any comments or suggestions are welcome

tonybase:
Definition should be consistent with Proto, and can be used as documentation or API metadata.

PR #1993, opened 2022-05-18
feat: Added ability to configure protojson

windzhu0514:
use kratos with watch apollo config.
add a new config item in apollo server,after publish ,kartos config watch works,but UnmarshalOptions.DiscardUnknown default is false,so scan will failed.

### DIFF
--- a/config/value.go
+++ b/config/value.go
@@ -1,15 +1,16 @@
 package config
 
 import (
-	"encoding/json"
+	stdjson "encoding/json"
 	"fmt"
 	"reflect"
 	"strconv"
 	"sync/atomic"
 	"time"
 
-	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
+
+	"github.com/go-kratos/kratos/v2/encoding/json"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


config/value.go
```diff
@@ -128,14 +129,14 @@ func (v *atomicValue) Duration() (time.Duration, error) {
...
+	data, err := stdjson.Marshal(v.Load())
 	if err != nil {
 		return err
 	}
 	if pb, ok := obj.(proto.Message); ok {
-		return protojson.Unmarshal(data, pb)
+		return json.UnmarshalOptions.Unmarshal(data, pb)
 	}
-	return json.Unmarshal(data, obj)
+	return stdjson.Unmarshal(data, obj)
```
tonybase, line 139: It is not possible to decode STD JSON directly, as fields may not match, It can be handled by two logic process.

windzhu0514: but proto json unmarshal should be can modify by user

tonybase:
```go
func (v *atomicValue) Scan(obj interface{}) error {
	codec := encoding. GetCodec(json.Name)
	data, err := codec.Marshal(v.Load())
	if err != nil {
		return err
	}
	return codec.Unmarshal(data, obj)
}

tonybase (approved):

PR #2001, opened 2022-05-20
missing defer

reason-tech:


tonybase:
This startup is synchronous

reason-tech:
> This startup is synchronous

As in #1982 said that the register must wait until all servers startup
if no `defer` exist, the `wg.Wait()` won't wait as `wg.Done()` finished immediately before the server start.

PR #2038, opened 2022-05-25
feat:build injection version

songzhibin97:
Use make all to automatically type version information into


shenqidebaozi:
Is the go install or go get installation applicable?

shenqidebaozi:
Note that it is a command line tool that needs to be pulled remotely

songzhibin97:
> Add more commits by pushing to the `[feat-version-Inject](/songzhibin97/kratos/tree/feat-version-Inject)` branch on [songzhibin97/kratos](/songzhibin97/kratos).

It won't work if you don't use makefile

Casper-Mars:
Hi~ Suppose a new tool is added, then the Makefile need to be modified. This requires extra work.

songzhibin97:
> Hi~ Suppose a new tool is added, then the Makefile need to be modified. This requires extra work.




> 
Yes, I'll turn it off if I don't need it, and the external version doesn't seem to be in use.

shenqidebaozi:
I don't think it makes sense

Issue #2037, opened 2022-05-24
[Question]  only use "WithSelector" without  "WithDiscovery" when New http client,  the selector does't work?


jursonmo:
```go
myselector := = wrr.New()
......
myselector.Apply(nodes)
client, err := http.NewClient(context.Background(), WithSelector(myselector))

client.Do(req) // selector does't work?
```

maybe the problem is here:
https://github.com/go-kratos/kratos/blob/70317d05a07a6382a1a18f14464771808c789e0c/transport/http/client.go#L272

should be :
```go
//if client.r != nil {
if client.opts.selector != nil {
```

PR #2039, opened 2022-05-25
fix: the selector doesn't work without Discovery

jursonmo:

the selector doesn't work without Discovery:
(transport/http): when create http client with WithSelector() without WithDiscovery(), the selector doesn't work when client.Do(req)

fixes: https://github.com/go-kratos/kratos/issues/2037


longXboy (dismissed):

tonybase (dismissed):

shenqidebaozi:
lint error

windzhu0514 (approved):

PR #2040, opened 2022-05-25
[WIP]feat(cmd/protoc-gen-go-kratos): generate kratos client

Casper-Mars:

#2010


cmd/protoc-gen-go-kratos/tests/library.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
```
shenqidebaozi, line 1: Should it be placed in the public test set？

PR #2044, opened 2022-05-26
feat(contrib/registry): support registry group

letian0805:

Support registrar group to register server to multi registry.

feat: #1940


tonybase:
contrib does not have go.mod, it is not recommended to add this.

PR #2090, opened 2022-06-09
fix(config): when watcher.Close() of config, the CPU load is very high

mrzxy:


haiyux (approved):

contrib/config/consul/watcher.go
```diff
@@ -56,7 +58,7 @@ func (w *watcher) Next() ([]*config.KeyValue, error) {
 	select {
 	case _, ok := <-w.ch:
 		if !ok {
-			return nil, nil
+			return nil, context.Canceled
```
Cluas, line 61: I think there is a problem with the logic itself, the Stop does not close the w.ch, so it never goes to the !ok branch

Cluas: FYI, We should define an error, probably called `ErrWatcherStopped`, and then after calling Stop, Next returns this error

mrzxy: > I think there is a problem with the logic itself, the Stop does not close the w.ch, so it never goes to the !ok branch

This happened because we called c.(*config.Config).Close()

Cluas: > 

https://github.com/go-kratos/kratos/blob/123fc1e6c81a92d0cb9512904382b4e2c3993653/contrib/config/consul/watcher.go#L67
watcher.Stop() only close closeChan

mrzxy: https://github.com/go-kratos/kratos/blob/752f011ba1beff0a69acda84acf73db349d0ef37/config/config.go#L64

c.Next() return nil when watcher.Stop(), It's going to loop forever

Cluas: The real reason is not that, I have fixed it, see this PR https://github.com/go-kratos/kratos/pull/2092

Cluas: cause by `case <-w.closeChan`

PR #2094, opened 2022-06-10
fix: kratos proto server cmd method name bugfix (#2089)

raw34:
- add ucFirst function and unit test at cmd/kratos/internal/proto/server/server.go
- fix the [method name style bug ](https://github.com/go-kratos/kratos/issues/2089)

### DIFF
--- a/cmd/kratos/internal/proto/server/server.go
+++ b/cmd/kratos/internal/proto/server/server.go
@@ -7,14 +7,9 @@ import (
 	"path"
 	"strings"
 
-	"golang.org/x/text/cases"
-	"golang.org/x/text/language"
-
 	"github.com/emicklei/proto"
 	"github.com/spf13/cobra"
 )
-
-var caser = cases.Title(language.English)
 
 // CmdServer the service command.
 var CmdServer = &cobra.Command{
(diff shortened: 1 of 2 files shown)


Casper-Mars:
This would not solve the issue #2089 . Issue #2089 is about camel case and snake case.

raw34:
> This would not solve the issue #2089 . Issue #2089 is about camel case and snake case.

this commit just fix the part2 problem, like this:


```
# proto
service OrderItems {
	rpc CreateOrderItems (CreateOrderItemsRequest) returns (CreateOrderItemsReply);
......
}
```

```
# wrong service
......

func (s *OrderItemsService) Createorderitems(ctx context.Context, req *pb.CreateOrderItemsRequest) (*pb.CreateOrderItemsReply, error) {
	return &pb.CreateOrderItemsReply{}, nil
}

.....
```

```
# right service

......

func (s *OrderItemsService) CreateOrderItems(ctx context.Context, req *pb.CreateOrderItemsRequest) (*pb.CreateOrderItemsReply, error) {
	return &pb.CreateOrderItemsReply{}, nil
}

.....
```

Casper-Mars:
OH~Do you mean this PR is still working in progress?

raw34:
> OH~Do you mean this PR is still working in progress?

the fact is i didn't find out the issue is about 2 problems, hah...

i can try to resolve the part1 problem some time later

shenqidebaozi:
@raw34 https://github.com/songzhibin97/kratos/blob/4e100fcdeb4dd046f1430ecde1ca96fafb9584ff/cmd/kratos/internal/proto/server/server.go#L70

shenqidebaozi:
It seems that the problem is caused by this modification

raw34:
> It seems that the problem is caused by this modification

yes, it is. the modification fix another bug, but trigger this bug. and my commit can fix it.

shenqidebaozi:
Has this PR been completed?

tonybase (approved):

haiyux (approved):

haiyux (approved):

PR #2106, opened 2022-06-15
feat: Add mux router option for http transport

nfangxu:

I use kratos as a lib, so I need to have a global mux.Router instead of using the built-in kratos. I added Option to achieve this. At the same time, I don't need the built-in Endpoint, so I added Option to disable it

### DIFF
--- a/transport/http/server.go
+++ b/transport/http/server.go
@@ -112,23 +112,36 @@ func Listener(lis net.Listener) ServerOption {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


shenqidebaozi:
You should first create a feature request issue to describe your idea.

nfangxu:
> For example, describe which methods have been added, how they are used, and whether incompatible modifications will be introduced.

Thanks for your reply, and I have created an issue to describe it

PR #2112, opened 2022-06-17
fix(binding): fix bind field mask

### LINKED ISSUE #2069: [protoc-gen-go-http] URL parameter fields dropped by generator for HTTP client
#### What happened:
When I define a API using protobuf with body specified, the extra fields (intended to be regarded as URL parameter) will be dropped in HTTP client code.

> it seems server side service is not affected and I can pass the extra parameters correctly.

All fields other than body field should be regarded as URL parameter to be keep align with the behavior of server side transporter.

Simply define a IDL for PATCH with body, then generate the client side code:

```protobuf
service UserService {
  rpc UpdateUser(UpdateUserRequest) returns (User) {
    option (google.api.http) = {
      patch: "/api/v1/users"
      body: "user"
    };
  }
}

message UpdateUserRequest {
  User user = 1;

  google.protobuf.FieldMask fields = 200;
}
```

In the IDL above, fields will be ignored when sending HTTP request to server side.

May change the [template code](https://github.com/go-kratos/kratos/blob/5de1f081f636cbaf503039eff0ff6ebb004d8337/cmd/protoc-gen-go-http/template.go#L81) to support it, or add a option in protoc-gen-go-http generator to enable this feature?

#### Environment:
- Kratos version (use `kratos -v`): v2.2.2 (I tried latest protoc-gen-go-http v2.3.0, same problem raises)
- Go version (use `go version`): go version go1.18.2 linux/amd64
- OS (e.g: `cat /etc/os-release`): Arch Linux
- Others: N/A

tonybase:
fix #2069


shenqidebaozi (approved):
LGTM

PR #2115, opened 2022-06-18
feat: zookeeper implements the config interface

pokitpeng:


pokitpeng:
This unit test must connect to ZooKeeper to pass 🙁

PR #2129, opened 2022-06-25
feat: support multi args for cmd proto add/server

deagon:


deagon:
Any reviews?

deagon:
Conflict resolved.

PR #2139, opened 2022-06-28
fix: remove the output of the proto command

shifengbin:
we input `kratos proto` in command line without any tips


haiyux (approved):
LGTM

shenqidebaozi (approved):
LGTM

Issue #2135, opened 2022-06-27
[Help Wanted] supplementary contrib unit test


shenqidebaozi:
To ensure the robustness of the plug-in package code, we need your help to supplement the unit test of contrib package.

- [ ] config 
- - [ ] apollo @lbbniu
- - [ ] consul @freezeChen
- - [ ] etcd @CodeFish-xiao  #2142
- - [x] kubernetes @rogerogers #2140
- - [x] nacos @wcvb13 #2145
- [ ] registry
- - [x] consul @freezeChen #2148
- - [ ] discovery @Juntaran
- - [ ] etcd @CodeFish-xiao
- - [ ] eureka @huangdi426
- - [x] kubernetes @rogerogers #2140
- - [x] nacos @wcvb13 #2145
- - [x] polaris @rogerogers #2196
- - [x] zookeeper @Ccheers  #2136
- [x] log
- - [x] aliyun @u5surf #2164
- - [x] fluent @u5surf #2141 
- - [x] zap @Ccheers #2136
- [ ] metrics @rogerogers
- [x] opensergo @zoujiejun #2137
- [x] encoding
- - [x] msgpack 

You can comment on the tasks you want to receive below, and I will assign them to you.

**Notice**: please use "part of" not "resolves" or "fixes", or it will close this original issue

CodeFish-xiao:
etcd

Ccheers:
zookeeper & zap

u5surf:
fluent
Can I use stretchr/testify?

zoujiejun:
encoding/msgpack

zoujiejun:
opensergo

wcvb13:
nacos

freezeChen:
consul

haiyux:
> fluent Can I use stretchr/testify?

Please don't use it。

rogerogers:
config.kubernetes

CodeFish-xiao:
[registry.etcd](https://github.com/go-kratos/kratos/pull/2142)

CodeFish-xiao:
> 

registry.etcd  https://github.com/go-kratos/kratos/pull/2142

rogerogers:
这个registry.k8s感觉测试有点困难，代码场景是在pod里面，那个getnamespace的文件跟podname都不好构造，用pod直接注册类似dubbo，跳过了k8s的健康检查吧

u5surf:
aliyun

rogerogers:
[...]

PR #2145, opened 2022-06-30
feat(contrib): add nacos unit test and remove unused function

wcvb13:
**Description (what this PR does / why we need it):**
add nacos unit test

**Which issue(s) this PR fixes (resolves / be part of):**
part of [#2135](https://github.com/go-kratos/kratos/issues/2135)

Other special notes for the reviewers:

### DIFF
--- a/contrib/config/nacos/config.go
+++ b/contrib/config/nacos/config.go
@@ -4,28 +4,18 @@ import (
 	"context"
 	"path/filepath"
 	"strings"
-	"time"
 
-	"github.com/go-kratos/kratos/v2/config"
 	"github.com/nacos-group/nacos-sdk-go/clients/config_client"
 	"github.com/nacos-group/nacos-sdk-go/vo"
+
+	"github.com/go-kratos/kratos/v2/config"
 )
 
 type Option func(*options)
 
 type options struct {
-	endpoint string // nolint:structcheck,unused
-
-	namespaceID string // nolint:structcheck,unused
-
 	group  string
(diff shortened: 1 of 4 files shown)


contrib/config/nacos/config.go
```diff
@@ -42,34 +32,6 @@ func WithDataID(dataID string) Option {
...
-	return func(o *options) {
-		o.logLevel = logLevel
-	}
-}
-
-// WithTimeout With nacos config timeout.
-func WithTimeout(time time.Duration) Option {
-	return func(o *options) {
-		o.timeoutMs = uint64(time.Milliseconds())
-	}
```
Cluas, line 70: This is the option api exposed to the user, bring a bit of your own thinking and don't rely on IDE hints.

Cluas: A better way is to add the relevant Option tests, which will make your linter happy.

wcvb13: This option is a parameter of the nacos client, the nacos client is created by the user and passed into newsource function, so this option is unused

wcvb13: If you want to use this option, it will need to re-implement the newsource method

Cluas: Sorry, I misunderstood, it turns out that the option defined here is not used at all😅

tonybase (approved):

shenqidebaozi (approved):

PR #2146, opened 2022-06-30
feat(contrib):add etcd watcher withKeysOnly option

liuyi0618:

read value from etcd twice if no such opOption


haiyux (approved):
LGTM

tonybase (approved):

PR #2159, opened 2022-07-06
feat(filter): Implement HTTP cache feature

### LINKED ISSUE #2144: [Feature] HTTP cache implementation as middleware
Please see the FAQ in our main README.md before submitting your issue.

### What problem is the feature used to solve?
Hello there,
An HTTP cache system would be great in Kratos to avoid the execution of unchanged response and set the correct HTTP headers.
I already worked on an implementation for other servers/reverse-proxies like Caddy, Træfik, Skipper, Gin, Echo, ... so I think the [Souin](https://github.com/darkweak/souin) integration in Kratos as an HTTP cache middleware would be easy. With that it will support the CDN validation/invalidation too, set default cache-control header, store in a distributed system using olric/etcd or in memory with badger/nutsdb.
The main question is the location of this middleware. 
- Only on the Kratos repository
- The stable HTTP cache release on the Kratos repository and the unstable/dev on my repository (like the caddy cache-handler does)
- Only on my repository
I think having an official HTTP cache system middleware in the Kratos repository could be a gain for it.

Cheers ✌️

### References

- [Souin supported plugins/middlewares](https://github.com/darkweak/souin/tree/master/plugins)

darkweak:

It implements an HTTP cache filter to be RFC-7234 compliant with the new RFC about the Cache-Status header. It can set headers for and invalidate CDN providers like Cloudflare, Akamai, Fastly,... This can decrease the number of requests handled by the upstream and delegate to the CDNs the responsability to serve the static content without having thoughts about which header they expect. It also supports two non-distributed storage providers (badger and nutsDB (probably implement roseDB/lotusDB in a near future)) and two distributed storage providers (olric and etcd).
It implements the coalescing system, that mean if multiple same requests with the same context are sent to the upstream, only one will be sent, temporize the others and use the same response for the waiting requests.
This can increase your server performances and reduce the latency to serve a response.

resolves #2144


tonybase:
I think this is a very simple HTTP middleware, but it seems to be very complicated to implement.

darkweak:
@tonybase In fact you just have to run httpcache.NewHTTPCacheFilter(httpcache.ParseConfiguration(config))

darkweak:
I made it to work with the Kratos configuration object, that's why the configuration file is really huge. The implementation cannot be simpler IMHO but it's transparent for the end users.

darkweak:
@tonybase Do you think we have to simplify more the filter? Like always configure the cache management through the Kratos configuration? The filter handle the GraphQL caching too btw and can expose an API to purge the cache either by regex or by Surrogate-Key (like ykeys/xkeys in Varnish).

darkweak:
I'll have to tag the new version of the HTTP cache on my side (maybe today) before merge that.

darkweak:
Ready to merge!

PR #2171, opened 2022-07-08
feat(metadata): Ignore service if the dependencies are not found

bakjos:
It skips the service from the service list (metadata call) instead of failing with error if one of the service's dependencies is not found

### DIFF
--- a/api/metadata/server.go
+++ b/api/metadata/server.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"sync"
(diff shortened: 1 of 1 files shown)


api/metadata/server.go
```diff
@@ -76,6 +76,10 @@ func (s *Server) load() error {
 			}
 			fdps, e := allDependency(fdp)
 			if e != nil {
+				if e == protoregistry.NotFound {
```
tonybase, line 79: It would be better to use errors.Is here

bakjos: Thanks @tonybase you're right, done

tonybase (approved):

PR #2191, opened 2022-07-12
refactor(contrib): use internal package func to get host and port

gaffeyQiu:
refactor(contrib): use internal package func to get host and port


gaffeyQiu:
I saw the contrib package have own go.mod , can‘t belong kratos repo, It's my mistake

PR #2199, opened 2022-07-15
feat: add message queue

toventang:
1. Added the interface defines of message queue;
2. Added Kafka cases implement message queue.


shenqidebaozi:
Sorry, we have no plans for the implementation of event interface for the time being

PR #2246, opened 2022-07-29
feat(logging): middleware/logging client add log for peer with addr

youzhixiaomutou:

It is hard to get real peer addr when using discovery with gRPC request.
I can only find `selector.FromPeerContext(ctx)` to touch and log peer addr from ctx.
I think it is userful for dev and debugging.

### DIFF
--- a/middleware/logging/logging.go
+++ b/middleware/logging/logging.go
@@ -8,6 +8,7 @@ import (
 	"github.com/go-kratos/kratos/v2/errors"
 	"github.com/go-kratos/kratos/v2/log"
 	"github.com/go-kratos/kratos/v2/middleware"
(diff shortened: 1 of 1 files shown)


middleware/logging/logging.go
```diff
@@ -72,6 +77,7 @@ func Client(logger log.Logger) middleware.Middleware {
 				"kind", "client",
 				"component", kind,
 				"operation", operation,
+				"addr", addr,
```
tonybase, line 80: maybe use node

youzhixiaomutou: node is an interface with no args. how to output all args?

youzhixiaomutou: > maybe use node

node is an interface with no args. how to output all args?

PR #2251, opened 2022-08-02
refactor(log) aliyun use the same interface name

Germiniku:

`contrib log aliyun` interface use the same interface name `Logger`.


contrib/log/aliyun/aliyun.go
```diff
@@ -157,10 +158,8 @@ func toString(v interface{}) string {
 		key = strconv.FormatInt(v, 10)
 	case uint64:
 		key = strconv.FormatUint(v, 10)
-	case string:
-		key = v
```
tonybase, line 161: Why did you delete this?

Germiniku: type `string` implements fmt.Stringer interface.

tonybase: I tested, type `string` does not implement the `fmt.Stringer` interface.

tonybase (approved):

PR #2256, opened 2022-08-04
fix: in case url or form bind param error should return BadRequest

shifengbin:
in case url or form bind param error should return BadRequest


haiyux (approved):
LGTM

tonybase (approved):

PR #2265, opened 2022-08-10
Global logger

realityone:

Global `GetLogger` return the inner Logger in global logger appliance, this will make inconsistent logger after `SetLogger` is called.

Return global logger appliance directly to wrap the inner logger, always make the all logger is same from `GetLogger`.


tonybase (approved):

SeniorPlayer:
```go
func (a *loggerAppliance) Log(level Level, keyvals ...interface{}) error {
	return a.Logger.Log(level, keyvals...)
}
```

This increases the call depth of global logger.

PR #2312, opened 2022-08-23
feat:prefer ipv4 than ipv6

hshe:

ip4 and ip6 all exist perfer ip4 

ip4 and ip6 all exist perfer select ip4


icylight:
confilicted with https://github.com/go-kratos/kratos/pull/2298

PR #2325, opened 2022-08-29
feat(registry): consul support re registry service

shenqidebaozi:


Windfarer:
registry is noun and register is verb, maybe you should name it as re-register?

PR #2344, opened 2022-08-31
feat(transport) add custom http headers before call client

toeydevelopment:
Description (what this PR does / why we need it):

let's our developers can pass or add headers before call the client code that Kratos generated.
For example others client that need to check Api-Key in the header but for now (as i am understand) it isn't supported yet.


toeydevelopment:
;)

PR #2352, opened 2022-09-02
fix: Log print duplicate key val

demoManito:
- [x] (De duplication of duplicate log key values)

- [x] fix issues: #2347

I don't think it is necessary to de duplicate data by using `func With(l Logger, kv ...interface{})`, which may cause the output to be inconsistent with expectations.
`func WithUniq(l Logger, key string, val interface{})` can be used to ensure that the key and val are not duplicated. Users can select this function according to their needs

我认为没有必要在使用 `With` 时消除重复数据，这可能会导致输出与预期不一致。
`WithUniq` 可用于确保 KEY 不重复，并且保证 KEY 对应的 VAL 是最新的，用户可以根据需要自行选择此功能。

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -3,7 +3,6 @@ package log
 import (
 	"context"
 	"log"
-	"reflect"
 )
 
 // DefaultLogger is default logger.
@@ -15,10 +14,11 @@ type Logger interface {
 }
 
 type logger struct {
-	logger    Logger
-	prefix    []interface{}
-	hasValuer bool
-	ctx       context.Context
+	logger     Logger
+	prefix     []interface{}
+	prefixUniq map[string]int
+	hasValuer  bool
+	ctx        context.Context
 }
 
 func (c *logger) Log(level Level, keyvals ...interface{}) error {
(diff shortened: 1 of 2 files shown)


log/log.go
```diff
@@ -27,12 +27,33 @@ func (c *logger) Log(level Level, keyvals ...interface{}) error {
...
 		return err
 	}
 	return nil
 }
 
+// uniqKeys Under the same key, the Val corresponding to the last key shall prevail
+func uniqKeys(kvs []interface{}) []interface{} {
+	uniqMap := make(map[interface{}]interface{})
+	keys := make([]interface{}, 0, len(kvs))
+	for i := 0; i < len(kvs); i++ {
```
icylight, line 40: for i:=0;i<len(kvs); i = i+2

log/log.go
```diff
@@ -27,12 +27,33 @@ func (c *logger) Log(level Level, keyvals ...interface{}) error {
...
+		if i&1 == 1 { // only the key is checked for repeatability
+			continue
+		}
```
icylight, line 43: 这里不用要

shenqidebaozi:
#2350 Repeat?

demoManito:
> [第2350章](https://github.com/go-kratos/kratos/pull/2350)重复？

yes, I read # 2350 It may not be right to handle it at the same time

hrygo:
> #2350 Repeat?

我的实现方式保持了原有顺序，相同的key，保留后来者。我个人认为这两点比较重要。

tonybase:
这个会导致log性能变差，不太建议再做一次对象

PR #2360, opened 2022-09-05
fix: log val err output

demoManito:

err output: `INFO ts=2022-09-05T12:56:11+08:00 caller=caller %!s(log.Valuer=0x10ea120)=key1 value1=KEYVALS UNPAIRED`
`%!s(log.Valuer=0x10ea120)` log.Valuer func is error translated


tonybase:
error: value1=KEYVALS UNPAIRED
This error requires handling the key value.

PR #2361, opened 2022-09-05
refactor:simplified regexp

xiaoliwang:


tonybase (approved):

cmd/kratos/internal/base/vcs_url.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 var (
-	scpSyntaxRe = regexp.MustCompile(`^([a-zA-Z0-9_]+)@([a-zA-Z0-9._-]+):(.*)$`)
+	scpSyntaxRe = regexp.MustCompile(`^(\w+)@([\w.-]+):(.*)$`)
```
Windfarer, line 11: this is a unnecessary refactoring, this matching regex is from 
https://github.com/golang/go/blob/master/src/cmd/go/internal/vcs/vcs.go#L256

Windfarer (changes requested):

cmd/kratos/internal/base/vcs_url.go
xiaoliwang: the change was applied in the golang part, should this part follow the same?

xiaoliwang: > this is a unnecessary refactoring, this matching regex is from 
> https://github.com/golang/go/blob/master/src/cmd/go/internal/vcs/vcs.go#L256
> It is enough clear and accurate

@Windfarer the change was applied in the golang part, should this part follow the same?

PR #2377, opened 2022-09-15
cleanup: remove fmt and errors new err

demoManito:

static err should use `errors.New`, the performance will be higher than `fmt.Errorf`

reference: https://github.com/uber-go/guide/blob/master/style.md#errors

### DIFF
--- a/contrib/registry/nacos/registry.go
+++ b/contrib/registry/nacos/registry.go
@@ -15,6 +15,8 @@ import (
 	"github.com/go-kratos/kratos/v2/registry"
 )
 
+var ErrServiceInstanceNameEmpty = errors.New("kratos/nacos: ServiceInstance.Name can not be empty")
+
 var (
 	_ registry.Registrar = (*Registry)(nil)
 	_ registry.Discovery = (*Registry)(nil)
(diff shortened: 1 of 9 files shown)


haiyux (approved):

contrib/registry/nacos/registry.go
```diff
@@ -82,7 +83,7 @@ func New(cli naming_client.INamingClient, opts ...Option) (r *Registry) {
 // Register the registration.
 func (r *Registry) Register(_ context.Context, si *registry.ServiceInstance) error {
 	if si.Name == "" {
-		return fmt.Errorf("kratos/nacos: serviceInstance.name can not be empty")
+		return errors.New("kratos/nacos: serviceInstance.name can not be empty")
```
Casper-Mars, line 86: Would it be better to declare as a variable?

demoManito: What does it mean to 'declare as a variable'? means `fmt.error` better?

Casper-Mars: a global variable such as "var Errxxxxxx = errors.New("xxxxx")".

demoManito: > a global variable such as "var Errxxxxxx = errors.New("xxxxx")".

modified @Casper-Mars

Casper-Mars (approved):

haiyux (approved):

PR #2397, opened 2022-09-22
fix(metadata): sort services by lexicographical order on ListServices

deagon:
It's out of order on the reply of ListServices, which is so complex for human reading.

image


deagon:
recheck

tonybase (approved):

PR #2398, opened 2022-09-22
fix(registry/etcd) continue retry when regist failed

ThereWeGo:

测试过程:
etcd作为register时.
1. 封掉etcd endpoints ,模拟网络故障. (iptables -I INPUT -s xxxx -j DROP)
2. regist 重试次数消耗完后会导致heartbeat 函数return.
3. 解封,模拟网络恢复. (iptables -D INPUT -s xxxx -j DROP)
4. 程序不会尝试再次向etcd注册. 服务丢失

### DIFF
--- a/contrib/registry/etcd/registry.go
+++ b/contrib/registry/etcd/registry.go
@@ -190,7 +190,10 @@ func (r *Registry) heartBeat(ctx context.Context, leaseID clientv3.LeaseID, key 
 			}
 			if _, ok := <-kac; !ok {
 				// retry failed
-				return
+				if ctx.Err() != nil {
+					return
+				}
+				continue
 			}
 		}
 


shenqidebaozi:
Whether backoff retry should be considered to prevent too many services from being registered at the same time after cluster recovery？

ThereWeGo:
> Whether backoff retry should be considered to prevent too many services from being registered at the same time after cluster recovery？

这个问题出现在我们服务集群经历一次断网, 
网络恢复后, 服务的注册信息丢失了, 
不得不对所有服务进行了重启, 重新向etcd注册.

https://github.com/ThereWeGo/kratos/blob/1d0f855dcdd414ec297d1b07f388eaece3ef609c/contrib/registry/etcd/registry.go#L191
```
				time.Sleep(time.Duration(retreat[rand.Intn(len(retreat))]) * time.Second)
			}
			if _, ok := <-kac; !ok {
				// retry failed
				if ctx.Err() != nil {
					return
				}
				continue
			}
		}

		select {
		case _, ok := <-kac:
			if !ok {
				if ctx.Err() != nil {
					// channel closed due to context cancel
					return
				}
				// need to retry registration
				curLeaseID = 0
				continue
			}
		case <-r.opts.ctx.Done():
			return
		}
```
简单些可以和上面一样,  在continue前加随机暂停缓解压力

shenqidebaozi:
Can you directly submit the backoff retry function to this merge request

PR #2402, opened 2022-09-24
test(middleware/metadata): add Test_hasPrefix

180909:

### DIFF
--- a/middleware/metadata/metadata_test.go
+++ b/middleware/metadata/metadata_test.go
@@ -137,3 +137,24 @@ func Test_WithPropagatedPrefix(t *testing.T) {
 		t.Error("The prefix must be overridden.")
 	}
 }
+
+func Test_hasPrefix(t *testing.T) {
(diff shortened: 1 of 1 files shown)


middleware/metadata/metadata_test.go
```diff
@@ -137,3 +137,24 @@ func Test_WithPropagatedPrefix(t *testing.T) {
 		t.Error("The prefix must be overridden.")
 	}
 }
+
+func Test_hasPrefix(t *testing.T) {
```
demoManito, line 141: TestHasPrefix

demoManito: Golang does not recommend this naming method, and the standard library does not have this naming method

shenqidebaozi:
fix conflicting files

180909:
I check the file, and the `TestOptions_hasPrefix` function already exists. So, i close this PR.

PR #2412, opened 2022-09-28
cleanup: using constants instead of literals

demoManito:

use http package methods


Cluas (approved):
LGTM

Cluas (approved):

PR #2481, opened 2022-10-29
feat:health check pkg

shifengbin:
Let's see  this idea works


haiyux:
@tonybase  please take a look

Issue #2479, opened 2022-10-28
[Proposal] service health status check


shenqidebaozi:
- https://www.amitph.com/custom-health-check-spring-boot-actuator/

haiyux:
https://github.com/grpc/grpc-go/blob/master/health/server.go

tonybase:
HTTP/gRPC -> Health -> Component[ ping: redis/mysql ]

haiyux:
@shifengbin

shifengbin:
here

tonybase:
```go
type Checker interface {
	Check(ctx context.Context) error
}
type CheckerFunc func() error
func (f CheckerFunc) Check() error {
	return f()
}

type Health struct {}
func (h *Health) Check(ctx context.Context) error {}
func (h *Health) Start(ctx context.Context) error {}
func (h *Health) Stop(ctx context.Context) error {}
func (h *Health) SetStatus(service string, status Status) {}
func (h *Health) ServeHTTP(w http.ResponseWriter, r *http.Request) {}
```
```

Darrenzzy:
https://github.com/go-kratos/kratos/issues/2479#issuecomment-1305004846

对func的处理简单了点，需要支持如 any 这样类型的返回参数，gomock设计中接口设计较完善，可以参考下。

perriea:
@tonybase @shenqidebaozi #2541 fix the problem right ? Often the http scope is not covered

shenqidebaozi:
> @tonybase @shenqidebaozi #2541 fix the problem right ? Often the http scope is not covered

this issue is the design of the planned kratos health check interface, which has no direct relationship with your modification.

PR #2482, opened 2022-10-31
add health check

tonybase:
#2479

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -1,0 +1,78 @@
+package health
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+)
+
+type Option func(*Health)
+
+type Status string
+
+const (
+	StatusUp   = Status("UP")
+	StatusDown = Status("DOWN")
+)
+
+type Result struct {
+	Status Status `json:"status"`
+	Error  string `json:"error"`
+}
+
+// Health .
+type Health struct {
+	liveness map[string]Checker
+	readness map[string]Checker
+}
+
+func New() *Health {
+	return &Health{}
+}
+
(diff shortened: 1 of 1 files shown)


shenqidebaozi:
动态控制健康检查结果是先不实现吗？还是通过其他方式实现

shenqidebaozi:
代码很清真，看起来只需要修改为通过 grpc/http 借口将健康结果暴露即可，另外在服务生命周期是不是也要控制健康检查结果？

health/health.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// CheckerFunc is an adapter type to allow the use of ordinary functions as
+// health checks.
+type CheckerFunc func() error
+
+// CheckHealth calls f().
+func (f CheckerFunc) CheckHealth() error {
+	return f()
+}
```
jcyamacho, line 39: 🤔 maybe `ctx` param is mising in the adapter

health/health.go
```diff
@@ -0,0 +1,90 @@
...
+		if err := c.CheckHealth(ctx); err != nil {
+			res.Details[n] = Detail{Status: StatusDown, Error: err.Error()}
+		} else {
+			res.Details[n] = Detail{Status: StatusUp}
+		}
+	}
+	return res
+}
+
+func (h *Health) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
jcyamacho, line 82: it's ok to not have an option for creating 2 different handlers (liveness and readiness) ??
or the idea is to create 2 different `Health` and add only one category to each one??

tonybase: I think there's only liveness, we only deal with application health, and components are unavailable and measured by metrics, logs.

PR #2484, opened 2022-11-02
fix(filter): filter out keyvals if keyFunc is provided

migmartri:

This PR fixes a bug where a `filterFunc` couldn't filter/override a keyval. This was a regression introduced [here](https://github.com/go-kratos/kratos/pull/1901/files).

The root cause was that to take into account the log prefix, we made a copy of the original keyvals and the filter was applied on such copy. This means that any filterFunc that was meant to override the original keyvals, was changing the copy instead.

This PR changes the code to apply a filter func in both the prefix and the log keyvals. The test cases have been also updated to cover all the scenarios.

Refs #1897

### DIFF
--- a/log/filter.go
+++ b/log/filter.go
@@ -64,20 +64,18 @@ func (f *Filter) Log(level Level, keyvals ...interface{}) error {
 	if level < f.level {
 		return nil
 	}
-	// fkv is used to provide a slice to contains both logger.prefix and keyvals for filter
-	var fkv []interface{}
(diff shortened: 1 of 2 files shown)


tonybase (approved):

shenqidebaozi (approved):

Cynicism-zx:
This pr only solves the problem of filtering, but does not solve the problem of missing trace_id in the request log 
this is latest pr for this issue #2570

Issue #2488, opened 2022-11-04
[BUG] cannot overwrite the default selector implementation


shenqidebaozi:
Cannot overwrite the default selector implementation

Because of the init order of go, the init here will be executed first. At this time, when the user sets the selector again, the implementation cannot be overwritten

https://github.com/go-kratos/kratos/blob/2a65502be27bca8eebb63a3e1cc34b758f3d94d3/transport/grpc/balancer.go#L22-L31

```
package init sort:
dependency： main -> a -> b 
execution order: b -> a -> main
```

PR #2489, opened 2022-11-04
fix(selector): set global do not work

Cluas:

fix https://github.com/go-kratos/kratos/issues/2488

### DIFF
--- a/selector/global.go
+++ b/selector/global.go
@@ -1,6 +1,12 @@
 package selector
 
-var globalSelector Builder
+var globalSelector = &wrapSelector{}
+
+// wrapSelector wrapped Selector.
+// help override global Selector implementation
+type wrapSelector struct {
+	Builder
+}
 
 // GlobalSelector returns global selector builder.
 func GlobalSelector() Builder {
@@ -9,5 +15,5 @@ func GlobalSelector() Builder {
 
 // SetGlobalSelector set global selector builder.
 func SetGlobalSelector(builder Builder) {
-	globalSelector = builder
+	globalSelector.Builder = builder
 }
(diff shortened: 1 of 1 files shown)


selector/global.go
```diff
@@ -2,12 +2,23 @@ package selector
 
 var globalSelector Builder
```
tonybase, line 3: var globalSelector  *wrapSelector

selector/global.go
```diff
@@ -2,12 +2,23 @@ package selector
...
 }
 
 // SetGlobalSelector set global selector builder.
 func SetGlobalSelector(builder Builder) {
-	globalSelector = builder
+	if globalSelector == nil {
+		globalSelector = &wrapSelector{builder}
+		return
+	}
+	globalSelector.(*wrapSelector).Builder = builder
```
tonybase, line 23: globalSelector.Builder = builder

selector/global.go
Cluas: It will conflict with https://github.com/go-kratos/kratos/blob/0f38511ea8303b42007b5bdf7b85ebd6fafc833f/transport/grpc/client.go#L27

tonybase: `var globalSelector = &wrapSelector{wrr.NewBuilder()}`

Can be initialized by default.

Cluas: In fact, we can't do it. import cycle not allowed

tonybase: wrr: init()

Cluas:
@tonybase @shenqidebaozi PTAL

shenqidebaozi (approved):

PR #2575, opened 2022-12-12
fix(kratos/registry/registry.go): Kratos's ServiceInstance does not i…

shima-park:
Kratos's ServiceInstance does not implement an Equal method for grpc's Attributes.

Description (what this PR does / why we need it):
It will help the grpc resolver to work correctly.
If we don't do this, the grpc resolver cannot correctly determine whether the node already exists, and every time the service discovery generates an event, it will establish a connection with all the nodes in the event and release the previously existing node even if it is in the new event exists in.

[resolver.go](https://github.com/go-kratos/kratos/blob/b3eff576ce0d5cbadf53ec2a9ab1232d45e5cef3/transport/grpc/resolver/discovery/resolver.go#L70) 
[balancer.go](https://github.com/grpc/grpc-go/blob/master/balancer/base/balancer.go#L106) 
[map.go](https://github.com/grpc/grpc-go/blob/master/resolver/map.go#L60) 
[resolver.go](https://github.com/grpc/grpc-go/blob/master/resolver/resolver.go#L143) 
[attributes.go](https://github.com/grpc/grpc-go/blob/master/attributes/attributes.go#L91)

Which issue(s) this PR fixes (resolves / be part of):

Other special notes for the reviewers:

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -1,6 +1,9 @@
 package registry
 
-import "context"
+import (
+	"context"
+	"sort"
+)
 
 // Registrar is service registrar.
 type Registrar interface {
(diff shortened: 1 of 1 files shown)


longXboy (approved):

shenqidebaozi (approved):
LGTM

PR #2612, opened 2023-01-10
test(config/file): fix the unit test TestFile being blocked on Windows

mstmdev:


config/file/file_test.go
```diff
@@ -114,11 +114,14 @@ func testWatchFile(t *testing.T, path string) {
 	if err != nil {
 		t.Error(err)
 	}
-	defer f.Close()
```
haiyux, line 117:
```go
defer func() {
        if err = f.Close(); err != nil {
		t.Fatalf("close file error(%v)", err)
	}
}()
```

haiyux: Is this better？

mstmdev: On Windows systems, if you modify a file and call `watch.Next()` before closing it, the call may block indefinitely. This is because the `fsnotify` package uses the `syscall.GetQueuedCompletionStatus` function, which might not return the correct result until the file handle is closed.

PR #2616, opened 2023-01-13
feat: log helper support sprint option

woniu317:


woniu317:
LGTM

Windfarer:
Why this modification? Could you provide some examples or scenario about it?

woniu317:
> Why this modification? Could you provide some examples or scenario about it?

We need mask struct by mask using struct tag. We can not get mask tag when use `log.Errorf("stu:%v", stud)`.  So we want to override sprint to get the mask tag.

`
package log

import (
	"bytes"
	"fmt"
	"reflect"
	"strings"
	"testing"
)

func TestMask(t *testing.T) {
	logger := NewMaskLogger(DefaultLogger)
	log := NewHelper(logger)
	stud := Stud{
		Mail: "yunpeng.yun@163.com",
	}

	// ERROR stu=y*****************m
	log.Errorw("stu", stud)
	// ERROR msg=stu:{yunpeng.yun@163.com}
	log.Errorf("stu:%v", stud)
}

// We need mask struct by tag. For example, we have a struct like this:
type Stud struct {
	Mail string `mask:"mail"`
}

func NewMaskLogger(l Logger) Logger {
	return &MaskLogger{l}
}

// mask: Mask struct by role using struct tag.
func mask(v interface{}) string {
	rv := reflect.ValueOf(v)
	rt := reflect.TypeOf(v)
	if rt.Kind() != reflect.Struct {
		// can not get tag
		return fmt.Sprint(v)
	}

	var ret bytes.Buffer

	for i := 0; i < rt.NumField(); i++ {
		ft := rt.Field(i)
		or := fmt.Sprint(rv.Field(i).String())
		r := ft.Tag.Get("mask")

		if r == "" {
			ret.WriteString(or)
			continue
		}
		// mask
		if len(or) > 2 {
			or = or[:1] + strings.Repeat("*", len(or)-2) + or[len(or)-1:]
		} else {
			or = "**"
		}
		ret.WriteString(or)
	}

	return ret.String()
}

func (m *MaskLogger) Log(level Level, keyvals ...interface{}) error {
	for i := 1; i < len(keyvals); i = i + 2 {
		keyvals[i] = mask(keyvals[i])
	}
	return m.l.Log(level, keyvals...)
}

type MaskLogger struct {
	l Logger
}

`

Windfarer (approved):
LGTM, and it would be better to open a PR for this feature on documentation

shenqidebaozi (approved):
LGTM

PR #2647, opened 2023-02-03
fix: do not re register after service logout

shenqidebaozi:

### DIFF
--- a/contrib/registry/consul/client.go
+++ b/contrib/registry/consul/client.go
@@ -205,32 +205,32 @@ func (c *Client) Register(_ context.Context, svc *registry.ServiceInstance, enab
 			defer ticker.Stop()
 			for {
 				select {
-				case <-ticker.C:
-					select {
-					case <-c.ctx.Done():
-						_ = c.cli.Agent().ServiceDeregister(svc.ID)
-						return
-					default:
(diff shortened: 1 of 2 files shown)


contrib/registry/consul/client.go
```diff
@@ -206,7 +207,7 @@ func (c *Client) Register(_ context.Context, svc *registry.ServiceInstance, enab
 				select {
 				case <-ticker.C:
 					err = c.cli.Agent().UpdateTTL("service:"+svc.ID, "pass", "pass")
-					if err != nil {
+					if err != nil && !errors.Is(c.ctx.Err(), context.Canceled) && !errors.Is(c.ctx.Err(), context.DeadlineExceeded) {
```
Cluas, line 210:
```go
Maybe we can do this:
for {
    select {
    case <-ticker.C:
        select {
        case <-c.ctx.Done():
            return
        default:
            err = c.cli.Agent().UpdateTTL("service:"+svc.ID, "pass", "pass")
            .....
        }
    case <-c.ctx.Done():
	return
    }
}

shenqidebaozi: fix, PTAL

Cluas (dismissed):

Cluas (approved):

longXboy (approved):

PR #2672, opened 2023-02-22
fix: package path to filepath

songzhibin97:
// Package filepath implements utility routines for manipulating filename paths
// in a way compatible with the target operating system-defined file paths.
//
// The filepath package uses either forward slashes or backslashes,
// depending on the operating system. To process paths such as URLs
// that always use forward slashes regardless of the operating
// system, see the path package.
// Package path implements utility routines for manipulating slash-separated
// paths.
//
// The path package should only be used for paths separated by forward
// slashes, such as the paths in URLs. This package does not deal with
// Windows paths with drive letters or backslashes; to manipulate
// operating system paths, use the path/filepath package.

Filepath is used to process file paths path is used to process url paths


haiyux (approved):

haiyux:
解决下冲突 修复一下lint

haiyux (approved):

shenqidebaozi (approved):

PR #2720, opened 2023-03-10
fix: When e is nil, the application crashes.

tx7do:

### DIFF
--- a/app.go
+++ b/app.go
@@ -177,7 +177,7 @@ func (a *App) buildInstance() (*registry.ServiceInstance, error) {
 		for _, srv := range a.opts.servers {
 			if r, ok := srv.(transport.Endpointer); ok {
 				e, err := r.Endpoint()
-				if err != nil {
+				if err != nil || e == nil {
 					return nil, err
 				}
 				endpoints = append(endpoints, e.String())


shenqidebaozi (approved):

app.go
```diff
@@ -177,7 +177,7 @@ func (a *App) buildInstance() (*registry.ServiceInstance, error) {
 		for _, srv := range a.opts.servers {
 			if r, ok := srv.(transport.Endpointer); ok {
 				e, err := r.Endpoint()
-				if err != nil {
+				if err != nil || e == nil {
 					return nil, err
```
shenqidebaozi, line 181: if e = nil, It looks like two nil will be returned

app.go
```diff
@@ -177,7 +177,7 @@ func (a *App) buildInstance() (*registry.ServiceInstance, error) {
 		for _, srv := range a.opts.servers {
 			if r, ok := srv.(transport.Endpointer); ok {
 				e, err := r.Endpoint()
-				if err != nil {
+				if err != nil || e == nil {
```
Cluas, line 180: Maybe we should check for `nil` pointers and then directly continue to skip invalid `*url.URL`.

PR #2743, opened 2023-03-17
feat(config): support set config merge used codec

shenqidebaozi:

### DIFF
--- a/config/reader.go
+++ b/config/reader.go
@@ -3,16 +3,13 @@ package config
 import (
 	"bytes"
 	"encoding/gob"
-	"encoding/json"
 	"fmt"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


config/reader.go
```diff
@@ -3,16 +3,13 @@ package config
...
 	"strings"
 	"sync"
 
-	"github.com/imdario/mergo"
-	"google.golang.org/protobuf/encoding/protojson"
-	"google.golang.org/protobuf/proto"
-
 	"github.com/go-kratos/kratos/v2/log"
+
+	"github.com/imdario/mergo"
```
demoManito, line 12: Should be placed above `"github.com/go-kratos/kratos/v2/log"`

haiyux (approved):

PR #2763, opened 2023-03-27
chore(contrib/registry/consul): repeated judgment `<-c.ctx.Done()` add notes

demoManito:


shenqidebaozi:
https://go101.org/article/channel-closing.html

shenqidebaozi:
```
// The try-receive operation here is to
				// try to exit the sender goroutine as
				// early as possible. Try-receive and
				// try-send select blocks are specially
				// optimized by the standard Go
				// compiler, so they are very efficient.
				select {
				case <- stopCh:
					return
				default:
				}
```

Issue #2778, opened 2023-04-09
[Feature] Register yml to codec encoder


cyb0225:
### Feature-request
是否考虑把 yml 也加入到 codec 中，与 yaml 使用同一个 codec.

shenqidebaozi:
you can directly help submit a PR

cyb0225:
Do we need to support toml? I can help to add.

shenqidebaozi:
you can put him in contrib

PR #2811, opened 2023-05-04
feat(encoding): support yml and toml codec

cyb0225:

- add yml codec in encoding module which is reuse the yaml codec struct.
- add toml codec in contrib/encoding.

#2778


contrib/encoding/toml/toml.go
shenqidebaozi, line 1: put it directly with yaml?

PR #2868, opened 2023-06-14
fix(config): references between different sources are consistent when watch

### LINKED ISSUE #2801: [Feature] Config update cascade for `Config.Watch()`
### Feature-request

Let's say we have 2 config sources:

1. Remote config system, under `remote` key
2. Local config file, under `local` key

Then we merge them like this:

```go
c := config.New(config.WithSource(
    remote,
    file.NewSource(path),
))
```

The config will be:

```yaml
remote:
  somePassword: xxxxxxx

local:
  redis:
    password: ${remote.somePassword}
```

The remote config could be updated anytime, we have to monitor changes using `config.Watch()`.

In this example, we are using `local.redis.password` for Redis initialization, not `remote.somePassword`, but we can't watch `local.redis.password` because when a referenced config value gets updated won't trigger update for any other config that includes it, so we have to watch `remote.somePassword` for now. It's unintuitive.

### Proposal

I think we can build a Relation Map when we first loaded all configs, then using this map to do update cascade (must avoid loop), enable developers to watch any config key they want.

But this problem could be further complicated if the updated config values include another config key(s) which they don't have before, in that case, we may have to update our Relation Map first.

I could help with this if it's approved.

zhan3333:

Fixed an issue where watch did not work when referencing configuration updates between multiple sources

fixes #2801

/


zhan3333:
@shenqidebaozi Hi, our project needs to use this function, can we speed up the schedule?

shenqidebaozi:
@zhan3333 ok

shenqidebaozi:
@zhan3333 please fix lint

zhan3333:
@shenqidebaozi the lint fix has been completed

zhan3333:
@shenqidebaozi ping

shenqidebaozi:
could you please provide more detailed background and modification instructions

zhan3333:
> could you please provide more detailed background and modification instructions

Of course @shenqidebaozi 

**Problem:** 

When a configuration is referenced across source config, updates to the referenced configuration cannot affect the referenced key.

How it works: config.load() will merge all source config into reader, and when config.next() gets a source configuration change, only that source KVS will be merged into reader.

Because the reference configuration is already processed to a specific value in reader, merge to reader does not trigger an update of the reference configuration.

For example:

source1:
```
{
    "key": "${key2}"
}
```

source2:
```
{
    "key2": "value"
}
```

After config.Load(), the structure in reader looks like this:

reader:
```
{
    "key": "value",
    "key2": "value"
}
```

When source2 key2 value is updated to newValue, the original config.watch() logic merges new kvs with reader to obtain the new reader value:

new reader
```
{
    "key": "value",
    "key2": "newValue"
}
```

Clearly, updates to key2 do not affect the value of the reference configuration key, causing the problem.

**Solution:**

When watch changes, Re-read the sources, this will make the 'key' get the reference configuration value '${key2}' again, and create a new reader to replace the config.reader to avoid the merge does not trigger updates when the reference.

zhan3333:
@shenqidebaozi ping

PR #2898, opened 2023-06-29
feat(cmd): set custom module name for command new

Twacqwq:
support set custom module name for command new

#2355


shenqidebaozi:
can we add some unit tests

Twacqwq:
> can we add some unit tests

I should do this, I'll submit it within the week

Twacqwq:
@shenqidebaozi submitted for review.please check it when you're avaliable. thx

cmd/kratos/internal/project/project.go
```diff
@@ -39,6 +40,7 @@ func init() {
 	CmdNew.Flags().StringVarP(&branch, "branch", "b", branch, "repo branch")
 	CmdNew.Flags().StringVarP(&timeout, "timeout", "t", timeout, "time out")
 	CmdNew.Flags().BoolVarP(&nomod, "nomod", "", nomod, "retain go mod")
+	CmdNew.Flags().StringVarP(&moduleName, "module-name", "m", moduleName, "module name")
```
windzhu0514, line 43: can use the args[0] as project name or module name.when args[0] is a module name,project name is the final path element of args[0]

Twacqwq: > can use the args[0] as project name or module name.when args[0] is a module name,project name is the final path element of args[0]

Currently, args[0] serves as both the project name, and the final element is used as the module name. Therefore, I think it is necessary to add an additional parameter to specify the module name, as users might create a project named /path/to/kratos1 and expect the module name to be kratos2. hope my understanding is correct.

PR #2922, opened 2023-07-20
feat: Added `Set()` and `Get()` for `Context`

flc1125:

In order to facilitate the implementation of context assignment and value.

**Example:**

```go
w := wrapper{
		router: testRouter,
		req:    &http.Request{},
		res:    nil,
		w:      responseWriter{},
}

w.Set("test", "test")
w.Get("test")
```


shenqidebaozi:
why not use it directly `context.WithValue()` or `ctx.Value`?

PR #2941, opened 2023-08-02
feat: Add a stack log to allow multiple logs to be logged simultaneously

flc1125:

I can do this when I need to report logs to multiple different logs.


flc1125:
**To discuss:**

This topic describes whether to reserve options for different loggers.

**Exampl:**

- Logs of different levels can be reported for different loggers, for example, A Logger reports Debug and B Logger reports Info.

PR #2963, opened 2023-08-25
fix: error handling in app afterStop

eastfisher:


shenqidebaozi (approved):
LGTM

app.go
```diff
@@ -139,10 +139,14 @@ func (a *App) Run() error {
 	if err = eg.Wait(); err != nil && !errors.Is(err, context.Canceled) {
 		return err
 	}
-	for _, fn := range a.opts.afterStop {
-		err = fn(sctx)
+
+	if len(a.opts.afterStop) != 0 {
+		for _, fn := range a.opts.afterStop {
+			err = fn(sctx)
```
Casper-Mars, line 145: ignore err and continue the loop? or should we use log output the err?

youzhixiaomutou:
What's the point of using `if len(a.opts.afterStop) != 0 {...}` before `for-range`?

demoManito:
`len(a.opts.afterStop) != 0` doesn't make sense?

PR #3017, opened 2023-10-07
feat: slog

songzhibin97:
adaptation of slog

### DIFF
--- a/contrib/log/slog/slog.go
+++ b/contrib/log/slog/slog.go
@@ -1,0 +1,40 @@
+package slog
+
+import (
+	"fmt"
+	"log/slog"
+
+	"github.com/go-kratos/kratos/v2/log"
+)
+
+var _ log.Logger = (*Logger)(nil)
+
+type Logger struct {
+	log *slog.Logger
+}
+
+func NewLogger(slog *slog.Logger) log.Logger {
+	return &Logger{slog}
+}
+
+func (l *Logger) Log(level log.Level, keyvals ...interface{}) error {
+	keylen := len(keyvals)
+	if keylen == 0 || keylen%2 != 0 {
+		l.log.Warn(fmt.Sprint("Keyvalues must appear in pairs: ", keyvals))
+		return nil
+	}
+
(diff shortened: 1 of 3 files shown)


kvii:
Shall we using slog.Handler internally instead of slog.Logger? I mean, it will cause caller path issue.

Code reference:

https://github.com/golang/go/blob/46ea4ab5cb87e9e5d443029f5f1a4bba012804d3/src/log/slog/logger.go#L241-L253

songzhibin97:
> Shall we using slog.Handler internally instead of slog.Logger? I mean, it will cause caller path issue. 感觉用 slog.Handler 更合适一点。要不然会出 call path 问题。
> 
> Code reference:
> 
> https://github.com/golang/go/blob/46ea4ab5cb87e9e5d443029f5f1a4bba012804d3/src/log/slog/logger.go#L241-L253

You're right. It's more like logger in kratos. It adds call stack.

contrib/log/slog/slog.go
```diff
@@ -1,40 +1,73 @@
...
 	if keylen == 0 || keylen%2 != 0 {
-		l.log.Warn(fmt.Sprint("Keyvalues must appear in pairs: ", keyvals))
+		l.Warn(fmt.Sprint("Keyvalues must appear in pairs: ", keyvals))
 		return nil
 	}
 
 	switch level {
 	case log.LevelDebug:
-		l.log.Debug("", keyvals...)
+		l.Debug("", keyvals...)
```
kvii, line 62: Consider to add a special case for DefaultMessageKey? So users can use `helper.Info("info")` to print "msg=info"(not msg='' msg=info).

kvii:
可能得等到 kratos 版本升到 1.21 之后，这个 PR 才有可能被合并吧……

songzhibin97:
> 可能得等到 kratos 版本升到 1.21 之后，这个 PR 才有可能被合并吧……

是的，也有可能不会合，内部讨论可能直接将kratos/l
og 平替过来

PR #3067, opened 2023-11-01
feat(transport/grpc): grpc client supports health check option

youzhixiaomutou:
Refer to https://github.com/go-kratos/kratos/pull/2736#issue-1625835510 grpc client healthCheck is force to enable.

But in https://github.com/grpc/grpc/blob/master/doc/health-checking.md#grpc-health-checking-protocol, the `Watch()` returns  a stream resp:
```proto
rpc Watch(HealthCheckRequest) returns (stream HealthCheckResponse);
```
In some special business usecase，I must use nginx or any other gateway to proxy the grpc request. But it is not support stream resp and it will always return write_timeout. It will take some error in my case like:

`rpc error: code = Unavailable desc = last connection error: connection active but received health check RPC error: rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR"`

The nginx plus has impl the health check with https://docs.nginx.com/nginx/admin-guide/load-balancer/grpc-health-check/.
I think it should support an option for user to choice if I want to use the client's health check. And with the https://github.com/go-kratos/kratos/pull/2736#issue-162583551 user should notice it may affects the forward compatibility.

In this pr, it adds a  function `WithHealthCheck()` to disable health check(default is true).

### DIFF
--- a/transport/grpc/client.go
+++ b/transport/grpc/client.go
@@ -103,6 +103,15 @@ func WithNodeFilter(filters ...selector.NodeFilter) ClientOption {
 	}
 }
 
+// WithHealthCheck with health check
+func WithHealthCheck(healthCheck bool) ClientOption {
+	return func(o *clientOptions) {
+		if !healthCheck {
+			o.healthCheckConfig = ""
+		}
+	}
+}
+
 // WithLogger with logger
 // Deprecated: use global logger instead.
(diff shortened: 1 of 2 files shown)


shenqidebaozi (approved):
LGTM

haiyux (approved):

PR #3069, opened 2023-11-06
fix(log): `WithContext()` changed the ctx field of the parent log.Filter

### LINKED ISSUE #3061: The trace_id between requests is getting mixed up
#### What happened:

The trace_id between requests is getting mixed up.

The `trace_id` between requests is independent.

#### How to reproduce it (as minimally and precisely as possible):

Use a `log.Filter` as the global logger 
and enable both `tracing.Server()` and `logging.Server()` middlewares.

this is cauased by `log.WithContext()`.

https://github.com/go-kratos/kratos/blob/6cdd81811e102620ec93b125ea6ed38f06a5e03c/log/log.go#L65

#### Environment:

- Kratos version (use `kratos -v`): v2.6.3
- Go version (use `go version`): 1.20

ionling:

Fix the concurrent problem of `log.WithContext()` function.

Fix #3061


shenqidebaozi:
Do we need to determine if `f.ctx=ctx` when `f.ctx=nil`? Can we also supplement UT to prove that the modifications have solved the problem and are compatible forward and backward

ionling:
I have added a UT for this.

About `f.ctx=ctx` when `f.ctx=nil`, I think setting a default value is a better solution.
As for backwards compatibility, I think it's a breakable change if there's any code that relies on it.

shenqidebaozi:
The submitted fix appears to have data race issues

ionling:
The data race problem was introduced by #2869.
I fixed it using a simpler solution.

shenqidebaozi:
please fix lint

haiyux (approved):

PR #3082, opened 2023-11-13
fix:polaris_config_watch

tan-zhuo:
the high CPU usage issue caused by watching the Polaris config, particularly in situations when config events are closed.

If the configuration connection has already been closed, the configuration file monitoring should be canceled.

### DIFF
--- a/contrib/polaris/config.go
+++ b/contrib/polaris/config.go
@@ -1,6 +1,7 @@
 package polaris
 
 import (
+	"context"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


contrib/polaris/config.go
```diff
@@ -110,7 +111,7 @@ func (w *ConfigWatcher) Next() ([]*config.KeyValue, error) {
 		}
 		return w.cfg, nil
 	}
-	return w.cfg, nil
+	return w.cfg, context.Canceled
```
shenqidebaozi, line 114: Is it reasonable to return a context error

tan-zhuo: I think it's reasonable because when the connection is already closed, it is indeed necessary to throw a context exception to the outer layer to cancel the file listener.

shenqidebaozi (dismissed):

haiyux (dismissed):

Issue #3110, opened 2023-12-08
errors had been ignore or covered when too much err happend


SchopenhauerZhang:
#### What happened:
when too much err happened ,error info has been ignored. 
由于err变量是同一个，后续的err会把之前执行过程中的error信息覆盖掉，导致错误信息被掩盖。
而且由于是FIFO的方式执行hook function；后续执行过程中不管有没有发生err，都会覆盖掉此前出现的err。
#### What you expected to happen:
avoid error info being ignore when too much err happened .
如果出现大量的err，在对变量重复赋值之前打印出err信息，将此前执行过程中的err暴露出来。
#### How to reproduce it (as minimally and precisely as possible):

````
for _, fn := range a.opts.afterStop {
		if err != nil {
			log.Warnf("%s", err)
		}

		fnErr := fn(sctx)
		if fnErr != nil {
			err = fnErr
		}
	}
````

````
for _, fn := range a.opts.beforeStop {
		if err != nil {
			log.Warnf("%s", err)
		}
		fnErr := fn(sctx)
		if fnErr != nil {
			err = fnErr
		}
	}
````
#### Environment:
- Kratos version (use `kratos -v`):lastest
- Go version (use `go version`):1.21.5
- OS (e.g: `cat /etc/os-release`):
- Others:

PR #3111, opened 2023-12-08
#3110  bugfix  instead throw errors  when too much err happend  not to cover errors

SchopenhauerZhang:
https://github.com/go-kratos/kratos/issues/3110

when too much errors happened ,print history err;
如果发生过error，就把发生过的error一并返回；而不是通过重复赋值将error =nil进行掩盖。

### DIFF
--- a/app.go
+++ b/app.go
@@ -143,7 +143,11 @@ func (a *App) Run() error {
 		if err != nil {
 			log.Warnf("%s", err)
 		}
-		err = fn(sctx)
+
+		fnErr := fn(sctx)
+		if fnErr != nil {
+			err = fnErr
+		}
 	}
 	return err
 }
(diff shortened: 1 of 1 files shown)


SchopenhauerZhang:
CR,Please,THX   -：)

SchopenhauerZhang:
CR,Please,THX   -：)

app.go
```diff
@@ -149,7 +156,13 @@ func (a *App) Run() error {
 func (a *App) Stop() (err error) {
 	sctx := NewContext(a.ctx, a)
 	for _, fn := range a.opts.beforeStop {
-		err = fn(sctx)
+		if err != nil {
```
shenqidebaozi, line 159: what is the purpose of this code?

SchopenhauerZhang: if err happened;but next fn has no err;this code will cover history err;
````
for {
// first time ;err != nil 
err = fn(sctx) 

// next time err = nil 
err = nil 
}
// return nil;err had been covered.
return err 

```

PR #3121, opened 2023-12-15
feat: use writer directly in stdLogger.

kvii:

log.stdLogger use log.Logger internally, and disable all options such as prefix or log flags. log.stdLogger use l.log like a io.Writer. So I removed l.log, use io.Writer directly.

Code "stdLogger use log.Logger with no options":
https://github.com/go-kratos/kratos/blob/08300d8a4178b4e0ff7b226887d66c0764f0d172/log/std.go#L20-L21

Code "stdLogger not use any format features of log.Logger":
https://github.com/go-kratos/kratos/blob/08300d8a4178b4e0ff7b226887d66c0764f0d172/log/std.go#L39-L43

I did not add test for stdLogger.Close, because there is no way to call Close outside.


kvii:
A benchmark for it.

* new code

```sh
goos: darwin
goarch: arm64
pkg: github.com/go-kratos/kratos/v2/log
BenchmarkNew-10     16720666            72.12 ns/op        0 B/op          0 allocs/op
PASS
```

* old code

```sh
goos: darwin
goarch: arm64
pkg: github.com/go-kratos/kratos/v2/log
BenchmarkOld-10      6824382           158.4 ns/op        32 B/op          1 allocs/op
PASS
```

shenqidebaozi (approved):

kvii:
@hawkingrei And could you help me to review this code too? This code removed unnecessary "log.Output"  so it doubles the performance of stdLogger.

daemon365 (approved):

PR #3131, opened 2023-12-19
feat: add server option for NotFoundHandler and MethodNotAllowedHandler

liubing0427:
pprof will be registered to DefaultServeMux by default, and go-kratos's NotFoundHandler and MethodNotAllowedHandler will use DefaultServeMux, resulting in the inability to flexibly control the switch of pprof. 
This PR adds ServerOption to modify NotFoundHandler and MethodNotAllowedHandler

### DIFF
--- a/transport/http/server.go
+++ b/transport/http/server.go
@@ -141,6 +141,18 @@ func PathPrefix(prefix string) ServerOption {
 	}
 }
 
+func NotFoundHandler(handler http.Handler) ServerOption {
+	return func(s *Server) {
+		s.router.NotFoundHandler = handler
+	}
+}
+
+func MethodNotAllowedHandler(handler http.Handler) ServerOption {
+	return func(s *Server) {
+		s.router.MethodNotAllowedHandler = handler
+	}
+}
+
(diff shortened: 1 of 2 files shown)


transport/http/server.go
```diff
@@ -177,12 +189,12 @@ func NewServer(opts ...ServerOption) *Server {
...
 		router:      mux.NewRouter(),
 	}
+	srv.router.NotFoundHandler = http.DefaultServeMux
+	srv.router.MethodNotAllowedHandler = http.DefaultServeMux
 	for _, o := range opts {
 		o(srv)
 	}
 	srv.router.StrictSlash(srv.strictSlash)
-	srv.router.NotFoundHandler = http.DefaultServeMux
-	srv.router.MethodNotAllowedHandler = http.DefaultServeMux
```
shenqidebaozi, line 185: what is the reason for adjusting the code order?

liubing0427: Move them before the server option settings so that they can be modified through the server option

shenqidebaozi (approved):

PR #3146, opened 2024-01-04
fix: clear err after app stop err handling

lawlielt:
clear the err after the app wait err handing, to avoid the err return when afterStop is not set.


shenqidebaozi (approved):
LGTM

Windfarer (approved):

PR #3147, opened 2024-01-04
fix: fix consul heartbeat 404 error

lawlielt:
this is will happen when heartbeat start after deregister, some error like below:
```
'[Consul] update ttl heartbeat to consul failed! err=Unexpected response code: 404 (Unknown check ID "xxxxxx". Ensure that the check ID is passed, not the check name.)
```


shenqidebaozi (approved):
LGTM

Windfarer (approved):

Issue #3141, opened 2023-12-28
EncodeURL will ignore some fields if request message contains repeated custom message


CaoYan1:
#### What happened:
I defined an RPC method:

```
message MsgInline {
}

message Request {
  string id = 1;
  repeated MsgInline msg = 2;
}

message Response {
}

service MyService {
  rpc Call(Request) returns (Response) {
    option (google.api.http) = {
      post: "/my-api/{id}",
      body: "*",
    };
  }
}
```

And I called it via http client, the id field in url path  was always empty.

#### Anything else we need to know?:
<img width="791" alt="image" src="https://github.com/go-kratos/kratos/assets/28662543/505aa117-14c8-4de7-9d6f-37cb4d6ce46c">

I find that form.EncodeValues(msg) will return error[unsupported message type:] if the request message contains repeated custom message, and ignore the rest fields, then my `id` field is ignored. I think an unsupported field shouldn't make other fields don't work.

#### Environment:
- Kratos version: v2.7.2
- Go version (use `go version`):  1.21.1
- OS (e.g: `cat /etc/os-release`): macos

CaoYan1:
My `Kratos` version is latest(2.7.2) and doesn't work still

kvii:
Can you provide your http client code? I try to reproduce your case, but there was no issue found.

Here is my reproduce [repo](https://github.com/kvii/issue3141).

```sh
# send http request
curl http://localhost:8000/my-api/1 -X POST -H 'Content-Type: application/json' -d '{"msg":[{}]}'

# console
id:"1"  msg:{}

CaoYan1:
Based on your repo, you can try this code. I can reproduce this issue.
'''
func TestHttpClient(t *testing.T) {
	client, _ := http.NewClient(context.Background(), http.WithEndpoint("localhost:8000"))
	c := NewMyServiceHTTPClient(client)
	resp, err := c.Call(context.Background(), &Request{
		Id: "id",
		Msg: []*MsgInline{
			{},
		},
	})
	fmt.Println(resp, err)
}
'''

kvii:
Oh, this is an issue of kratos http client, not server. I pushed a commit to my repo, add tests for this issue at [here](https://github.com/kvii/issue3141/blob/main/internal/data/client_test.go).

Source codes:

* Generated http client side code in my reproduce repo:

[...]

PR #3161, opened 2024-01-19
Fix issue #3141:修复post请求下无法一起解析query和数组body的bug

XiaoK29:

修复post请求下无法一起解析query和数组body的bug

修复#3141提出的bug

### DIFF
--- a/encoding/form/proto_encode.go
+++ b/encoding/form/proto_encode.go
@@ -22,6 +22,9 @@ func EncodeValues(msg interface{}) (url.Values, error) {
 		u := make(url.Values)
 		err := encodeByField(u, "", v.ProtoReflect())
 		if err != nil {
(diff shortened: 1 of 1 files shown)


encoding/form/proto_encode.go
```diff
@@ -22,6 +22,9 @@ func EncodeValues(msg interface{}) (url.Values, error) {
 		u := make(url.Values)
 		err := encodeByField(u, "", v.ProtoReflect())
 		if err != nil {
+			if len(u) > 0 {
```
shenqidebaozi, line 25: I hope to add some one-sided ones for this, do you think?

XiaoK29: 今天跑了一下例子，好像已经修复了

PR #3162, opened 2024-01-22
fix(consul): unable to recreate after canceled the consul watch

harbourlga:

### DIFF
--- a/contrib/registry/consul/registry.go
+++ b/contrib/registry/consul/registry.go
@@ -190,6 +190,7 @@ func (r *Registry) Watch(ctx context.Context, name string) (registry.Watcher, er
 			services:    &atomic.Value{},
 			serviceName: name,
 		}
+		set.ctx, set.cancel = context.WithCancel(context.Background())
 		r.registry[name] = set
 	}
 
@@ -208,10 +209,8 @@ func (r *Registry) Watch(ctx context.Context, name string) (registry.Watcher, er
 		// otherwise the initial data may be blocked forever during the watch.
 		w.event <- struct{}{}
 	}
-
 	if !ok {
-		err := r.resolve(ctx, set)
-		if err != nil {
+		if err := r.resolve(set.ctx, set); err != nil {
 			return nil, err
 		}
 	}
@@ -249,6 +248,9 @@ func (r *Registry) resolve(ctx context.Context, ss *serviceSet) error {
 				}
 				idx = tmpIdx
 			case <-ctx.Done():
+				r.lock.Lock()
+				delete(r.registry, ss.serviceName)
+				r.lock.Unlock()
 				return
 			}
 		}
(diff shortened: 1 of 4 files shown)


harbourlga:
agree

contrib/registry/consul/registry.go
```diff
@@ -209,12 +209,11 @@ func (r *Registry) Watch(ctx context.Context, name string) (registry.Watcher, er
...
-	if !ok {
-		err := r.resolve(ctx, set)
-		if err != nil {
-			return nil, err
-		}
+	err := r.resolve(ctx, set)
+	if err != nil {
+		return nil, err
 	}
+
```
shenqidebaozi, line 216:
```go
if err := r.resolve(ctx, set); err != nil {
     return nil, err
}
```

harbourlga: 已修改

shenqidebaozi:
Can it be fixed when multiple clients are created for the same service? For example, if a client is idle, will b client also be unable to update the instance due to Watcher shutdown?

harbourlga:
> Can it be fixed when multiple clients are created for the same service? For example, if a client is idle, will b client also be unable to update the instance due to Watcher shutdown?
> 
> https://github.com/go-kratos/kratos/blob/0cb74ea2a330b1a91ca892a158a041d8925becad/contrib/registry/consul/watcher.go#L34

If you are talking about multiple grpc clients, then each client will build a different resolver to watch name resolution updates. Each resolver has its own context to control the shutdown of the Watcher

https://github.com/grpc/grpc-go/blob/02858ee5064040458f7e3e04e10765a62814c50b/resolver_wrapper.go#L80

https://github.com/go-kratos/kratos/blob/0cb74ea2a330b1a91ca892a158a041d8925becad/transport/grpc/resolver/discovery/builder.go#L85

https://github.com/go-kratos/kratos/blob/0cb74ea2a330b1a91ca892a158a041d8925becad/transport/grpc/resolver/discovery/builder.go#L88

harbourlga:
> Can it be fixed when multiple clients are created for the same service? For example, if a client is idle, will b client also be unable to update the instance due to Watcher shutdown?
> 
> https://github.com/go-kratos/kratos/blob/0cb74ea2a330b1a91ca892a158a041d8925becad/contrib/registry/consul/watcher.go#L34

If you are talking about multiple grpc clients, then each client will build a different resolver to watch name resolution updates. Each resolver has its own context to control the shutdown of the Watcher. Therefore, when client a becomes idle, client b will not close the watcher.

When creating the grpc client, a resolver will be built: 
https://github.com/grpc/grpc-go/blob/02858ee5064040458f7e3e04e10765a62814c50b/resolver_wrapper.go#L80
https://github.com/go-kratos/kratos/blob/0cb74ea2a330b1a91ca892a158a041d8925becad/transport/grpc/resolver/discovery/builder.go#L86

https://github.com/go-kratos/kratos/blob/0cb74ea2a330b1a91ca892a158a041d8925becad/transport/grpc/resolver/discovery/builder.go#L88

shenqidebaozi (approved):

daemon365 (approved):

PR #3184, opened 2024-01-31
feat: ignore message parse error

### LINKED ISSUE #3141: EncodeURL will ignore some fields if request message contains repeated custom message

kvii:

Ignore proto message field errors when parseing path template.

fixes/resolves #3141

Unstable PR. Do not merge this PR to quickly.

### DIFF
--- a/encoding/form/proto_encode.go
+++ b/encoding/form/proto_encode.go
@@ -20,7 +20,7 @@ func EncodeValues(msg interface{}) (url.Values, error) {
 	}
 	if v, ok := msg.(proto.Message); ok {
 		u := make(url.Values)
-		err := encodeByField(u, "", v.ProtoReflect())
+		err := encodeByField(u, "", v.ProtoReflect(), false)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


xxsoul:
issue3141 (https://github.com/go-kratos/kratos/issues/3141) The root cause of EncodeURL ignoring fields is that its internal encodeMessage function does not support encoding Message-type data and returns err, and the current revision process destroys the atomicity of the encodeByField function's parsing of the Message, so it may not be appropriate. 
To fix this issue, consider extending the encodeMessage function to support encoding Message-type data or explicitly restricting the use of arrays of custom datatypes in this scenario through documentation.

kvii:
@xxsoul It just a demo. I don't want this commit to be merged.

PR #3199, opened 2024-02-18
fix(registry): goroutine leak of etcd watcher when ctx is done

ionling:

The watcher will call the `reWatch()` method when the ctx is done,
which is unintended and will result in a goroutine leak.

See https://github.com/etcd-io/etcd/blob/5d45a88ab73cc5461ccf23e4df70598d8d433ff0/client/v3/watch.go#L65:

> If the context "ctx" is canceled or timed out, returned "WatchChan" is closed,
> and "WatchResponse" from this closed channel has zero events and nil "Err()".

### DIFF
--- a/contrib/registry/etcd/watcher.go
+++ b/contrib/registry/etcd/watcher.go
@@ -6,6 +6,7 @@ import (
 
 	clientv3 "go.etcd.io/etcd/client/v3"
 
+	"github.com/go-kratos/kratos/v2/log"
 	"github.com/go-kratos/kratos/v2/registry"
 )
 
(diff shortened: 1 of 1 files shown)


contrib/registry/etcd/watcher.go
```diff
@@ -52,7 +53,16 @@ func (w *watcher) Next() ([]*registry.ServiceInstance, error) {
 	case <-w.ctx.Done():
 		return nil, w.ctx.Err()
 	case watchResp, ok := <-w.watchChan:
-		if !ok || watchResp.Err() != nil {
+		// Refer https://github.com/etcd-io/etcd/blob/5d45a88ab73cc5461ccf23e4df70598d8d433ff0/client/v3/watch.go#L65
+		if !ok && watchResp.Err() == nil {
+			// ctx is canceled or timed out
+			return nil, w.ctx.Err()
+		}
+		if watchResp.Err() != nil {
```
shenqidebaozi, line 61: What if there are other errors in handling this here?

shenqidebaozi: Do you want to recreate it as long as there are errors? If it's to solve the problem of idle time? gRPC will actively call to create a watch, and supports re creation.

ionling: The `reWatch` logic was written by someone else, and I haven't modified it.

> Do you want to recreate it as long as there are errors? If it's to solve the problem of idle time? gRPC will actively call to create a watch, and supports re creation.

ionling: Based on the documentation for `Watch`, I believe it may help resolve the `ErrCompacted` error.

> If revisions waiting to be sent over the watch are compacted, 
> then the watch will be canceled by the server, 
> the client will post a compacted error watch response, and the channel will close.

PR #3234, opened 2024-03-12
fix(encoding/form): optimize EncodeField and add test cases

demoManito:

- remove protoreflect.StringKind convert to string
image

- supplementary encode map test cases


encoding/form/proto_encode.go
```diff
@@ -137,14 +137,12 @@ func EncodeField(fieldDescriptor protoreflect.FieldDescriptor, value protoreflec
...
 		return string(desc.Name()), nil
-	case protoreflect.StringKind:
-		return value.String(), nil
 	case protoreflect.BytesKind:
 		return base64.URLEncoding.EncodeToString(value.Bytes()), nil
 	case protoreflect.MessageKind, protoreflect.GroupKind:
 		return encodeMessage(fieldDescriptor.Message(), value)
 	default:
-		return fmt.Sprint(value.Interface()), nil
+		return value.String(), nil
```
demoManito, line 145: image

shenqidebaozi, line 145: Should this be a feature or fix?

demoManito: image

`value.String()` already contains `fmt.Sprint(value.Interface())`, for non-special types, just use `.String()`, optimized by the way.

shenqidebaozi: What I'm talking about is the title issue

demoManito: done @shenqidebaozi

Issue #3181, opened 2024-01-31
Test failed on different time zone.


kvii:
#### What happened:

Test failed. Logs like below:

```sh
go test -run ^TestEncodeValues$ github.com/go-kratos/kratos/v2/encoding/form

--- FAIL: TestEncodeValues (0.00s)
    /Users/kvii/workspace/github.com/kvii/kratos/encoding/form/proto_encode_test.go:49: want: a=19...&timestamp=1970-01-01T00%3A00%3A20.000000002Z..., got: a=19...&timestamp=1970-01-01T08%3A00%3A20.000000002Z...
FAIL
FAIL	github.com/go-kratos/kratos/v2/encoding/form	1.340s
FAIL
```

Note that timestamps "1970-01-01**T00**" and "1970-01-01**T08**" are different.

#### What you expected to happen:

Test passed.

#### How to reproduce it (as minimally and precisely as possible):

`go test -run ^TestEncodeValues$ github.com/go-kratos/kratos/v2/encoding/form`.

#### Anything else we need to know?:

I'll fix it soon.

#### Environment:
- Kratos version (use `kratos -v`): v2.7.2
- Go version (use `go version`): go1.21.6 darwin/arm64
- OS (e.g: `cat /etc/os-release`): Mac OS Sonoma 14.3
- Others:

PR #3257, opened 2024-03-17
fix(encoding/form): `Timestamp` use local time and fix test wrong mock time zone

demoManito:

1. The mock time zone should not be UTC. The time zone should be fixed when passing parameters.
2. decode `parseMessage` should also use the current time zone time

fix issue: https://github.com/go-kratos/kratos/issues/3181

### DIFF
--- a/encoding/form/proto_decode.go
+++ b/encoding/form/proto_decode.go
@@ -244,7 +244,7 @@ func parseMessage(md protoreflect.MessageDescriptor, value string) (protoreflect
 		if value == nullStr {
 			break
 		}
-		t, err := time.Parse(time.RFC3339Nano, value)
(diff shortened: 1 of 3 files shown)


encoding/form/proto_decode.go
```diff
@@ -244,7 +244,7 @@ func parseMessage(md protoreflect.MessageDescriptor, value string) (protoreflect
 		if value == nullStr {
 			break
 		}
-		t, err := time.Parse(time.RFC3339Nano, value)
+		t, err := time.ParseInLocation(time.RFC3339Nano, value, time.Local)
```
demoManito, line 247: image
Synchronize time zone with encoding timestamp

shenqidebaozi (approved):

PR #3279, opened 2024-04-15
feat(transport): expose server implementation on transport

NathanBaulch:

Expose the server implementation on `Transporter`, similar to what's available in GRPC interceptors. This gives middlewares the ability to make decisions or delegate functionality to the server/handlers that do the work.

Example middleware that delegates authentication checks to the server handler:

```golang
func Server() middleware.Middleware {
	return func(handler middleware.Handler) middleware.Handler {
		return func(ctx context.Context, req any) (any, error) {
			if tr, ok := transport.FromServerContext(ctx); !ok {
				return nil, errors.Unauthorized()
			} else if auth, ok := tr.Server().(Authenticator); !ok {
				return nil, errors.Unauthorized()
			} else if ctx, err := auth.Authenticate(ctx, tr); err != nil {
				return nil, err
			} else {
				return handler(ctx, req)
			}
		}
	}
}
```

This is an additive change that requires HTTP servers to be regenerated since they need to call `http.SetServer` alongside the existing `http.SetOperation`.

### DIFF
--- a/app.go
+++ b/app.go
@@ -102,7 +102,7 @@ func (a *App) Run() error {
 		server := srv
 		eg.Go(func() error {
 			<-ctx.Done() // wait for stop signal
-			stopCtx := octx
+			stopCtx := context.WithoutCancel(octx)
 			if a.opts.stopTimeout > 0 {
 				var cancel context.CancelFunc
 				stopCtx, cancel = context.WithTimeout(stopCtx, a.opts.stopTimeout)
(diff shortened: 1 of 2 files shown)


shenqidebaozi:
This change looks strange, I don't know why it's done this way

NathanBaulch:
I'm more than happy to answer any questions. My approach exactly mirrors the way `Transport.Operation` (aka grpc `FullMethod`) is intercepted and exposed to middlewares.

NathanBaulch:
For now I'm forced to work around this by post-processing the generated code with:

```shell
# hack: ensure http handlers capture server
sed -i 's/http.SetOperation/http.SetServer(ctx, srv)\n\t\t&/g' ../internal/api/server/v1/*_http.pb.go
```

Issue #3035, opened 2023-10-13
load apollo config yml format panic


hoslo:
#### What happened:
panic when load apollo yml config
#### What you expected to happen:
do not panic
#### How to reproduce it (as minimally and precisely as possible):
```golang
	type Project struct {
		Test int `yaml:"test" json:"test" yml:"test"`
	}
	type Bootstrap struct {
		Ns      string   `yaml:"ns" json:"ns" yml:"name"`
		Name    string   `yaml:"name" json:"name" yml:"name"`
		Project *Project `yaml:"project" json:"project" yml:"project"`
	}
	c := config.New(
		config.WithSource(
			apollo.NewSource(
				apollo.WithAppID("kratos"),
				apollo.WithCluster("dev"),
				apollo.WithEndpoint("http://localhost:8080"),
				apollo.WithNamespace("app.yml"),
				apollo.WithEnableBackup(),
				// apollo.WithOriginalConfig(),
				apollo.WithSecret("37b75fb736084e41b065fd11233f70e2"),
			),
		),
	)

	if err := c.Load(); err != nil {
		panic(err)
	}
```
#### Environment:
- Kratos version (use `kratos -v`): kratos version v2.7.0
- Go version (use `go version`): go version go1.21.1 darwin/arm64
- OS (e.g: `cat /etc/os-release`): macos
- Others:

shenqidebaozi:
I don't think this is a bug, yaml or yml is just a difference in declaring suffixes. If we add a separate yml encoding package for yml, I think it would be strange

hoslo:
> I don't think this is a bug, yaml or yml is just a difference in declaring suffixes. If we add a separate yml encoding package for yml, I think it would be strange

how you think should to solve this question?

shenqidebaozi:
The two codes are exactly the same, only the package with the same name does not look very halal 😂

hoslo:
> The two codes are exactly the same, only the package with the same name does not look very halal 😂

maybe we should put them in one package (yaml)?

PR #3304, opened 2024-04-25
fix: yaml.Codec supports both the keys "yaml" and "yml" at the same time

DCjanus:

In the past, when we tried to execute `./server -conf ./configs/config.yml`, it would generate an error message saying `failed to merge config source: unsupported key: config.yml format: yml`, but using `config.yaml` did not. 

This may be confusing because both `yaml` and `yml` are valid suffixes for `YAML`.

### DIFF
--- a/encoding/yaml/yml.go
+++ b/encoding/yaml/yml.go
@@ -1,0 +1,19 @@
+package yaml
+
+import (
+	"github.com/go-kratos/kratos/v2/encoding"
+)
+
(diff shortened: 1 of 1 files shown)


DCjanus:
@tonybase would you mind to take a look, which would be really helpful.

demoManito:
https://github.com/go-kratos/kratos/issues/3035#issuecomment-1760715296

PR #3311, opened 2024-04-30
feat(transport/http/binding): controls whether to use proto text name…

demoManito:
… as the key of the parameter

- `binding.EncodeURL` and `form.EncodeValues` and `form.EncodeFieldMask` add `option.EncodeOption`, used to control whether to use proto text name as key when encoding.
- `cmd/protoc-gen-go-http` add the `-jsonkey` parameter to control whether to use the json tag defined in proto as the key when encoding. The default is true and will not change the original generated code logic.

fixes: https://github.com/go-kratos/kratos/pull/2774#issuecomment-1518982578


demoManito:
@shenqidebaozi PTAL

demoManito:
Please take a look at this if you have time @shenqidebaozi. For users, there will be deviations in the use of EncodeURL. There are already multiple issues that have feedback on this.

encoding/form/option/encode.go
shenqidebaozi, line 1: Why do we need to separate this package? It looks strange to see it here

demoManito: It's true that a single package doesn't make much sense, it's already moved out @shenqidebaozi

shenqidebaozi:
Is it necessary to process in encoding and can it be processed in gen code

demoManito:
> Is it necessary to process in encoding and can it be processed in gen code是否需要在编码中处理，是否可以在 gen code 中处理

What does `gen code` mean? Since the code specifies using json name as a parameter, it seems that only this method will not affect the previous compatibility @shenqidebaozi

PR #3320, opened 2024-05-10
fix(contrib/registry/nacos): Abnormal blocking of the Subscribe and Next methods.

guoquanwei:
This submission can solve the channel blocking problem during service discovery, as well as the channel message missing problem when the GRPC idle state changes. ﻿

The first scenario is when both the configuration center and registration center are initialized at the same time during service startup, and the SubscribeCallback method will be triggered twice, throwing the error "over time discovering the creation of observers", resulting in a discovery failure. The second scenario occurs after NotLoadCacheAtStart: true. When grpc exits the idle state, calling newWatcher again will not trigger the SubscribeCallback method, causing the Next function to be blocked and the service to be unable to communicate.

### DIFF
--- a/contrib/registry/nacos/watcher.go
+++ b/contrib/registry/nacos/watcher.go
@@ -41,14 +41,16 @@ func newWatcher(ctx context.Context, cli naming_client.INamingClient, serviceNam
 		Clusters:    clusters,
 		GroupName:   groupName,
 		SubscribeCallback: func(services []model.SubscribeService, err error) {
-			if len(w.watchChan) == 0 {
-				w.watchChan <- struct{}{}
+			select {
+			case w.watchChan <- struct{}{}:
+			default:
 			}
 		},
 	}
 	e := w.cli.Subscribe(w.subscribeParam)
-	if len(w.watchChan) == 0 {
-		w.watchChan <- struct{}{}
+	select {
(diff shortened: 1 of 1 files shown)


contrib/registry/nacos/watcher.go
```diff
@@ -41,10 +41,15 @@ func newWatcher(ctx context.Context, cli naming_client.INamingClient, serviceNam
 		Clusters:    clusters,
 		GroupName:   groupName,
 		SubscribeCallback: func(services []model.SubscribeService, err error) {
-			w.watchChan <- struct{}{}
+			if len(w.watchChan) == 0 {
+				w.watchChan <- struct{}{}
+			}
```
demoManito, line 46:
```go
		select {
		case w.watchChan <- struct{}{}:
		default:
		}
```
It seems that using `select` would be a better choice

guoquanwei: The effect is the same, I can accept it all. Do you need me to modify it?

guoquanwei: I have switched to the recommended implementation and passed the test locally

demoManito (approved):

shenqidebaozi (approved):

PR #3357, opened 2024-07-03
feat(contrib/log/zap): If disable this level of logging entirely, av…

harbourlga:
* The zap log plug-in implemented for contrib/log currently filters unnecessary log levels after the log information is formatted, which will cause a lot of additional overhead.
* Referring to the implementation of SugaredLogger, it will first determine whether the log level meets the current needs before formatting the string.

None

None

### DIFF
--- a/contrib/log/zap/zap.go
+++ b/contrib/log/zap/zap.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 
 	"github.com/go-kratos/kratos/v2/log"
 )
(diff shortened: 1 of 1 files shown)


shenqidebaozi (dismissed):

contrib/log/zap/zap.go
```diff
@@ -3,9 +3,10 @@ package zap
 import (
 	"fmt"
 
-	"go.uber.org/zap"
-
 	"github.com/go-kratos/kratos/v2/log"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
```
demoManito, line 9: third-party packages should be placed before `github.com/go-kratos`

demoManito (approved):

shenqidebaozi (approved):

Issue #3361, opened 2024-07-09
使用北极星作为服务发现时，在注册多个grpc client时，启动报错：concurrent map iteration and map write


JackChanw:
#### What happened:
INFO msg=watcher's ctx cancel : context canceled
INFO msg=[resolver] update instances: [{"id":"7e55bf0237ad7a213c338824f08a89552fa2a1fb","name":"bv_media_server.grpc","version":"","metadata":{"env":"prod","env_id":"0","env_name":"","env_tag":"","protocol":"grpc","version":"","weight":"100"},"endpoints":["grpc://golden-us04.domob.cn:9789"]},{"id":"d742c26c247bbc377f8a99ac9c10e402f04da209","name":"bv_media_server.grpc","version":"","metadata":{"env":"prod","env_id":"0","env_name":"","env_tag":"","protocol":"grpc","version":"","weight":"100"},"endpoints":["grpc://golden-us03.domob.cn:9789"]}]
INFO msg=[resolver] update instances: [{"id":"7e55bf0237ad7a213c338824f08a89552fa2a1fb","name":"bv_media_server.grpc","version":"","metadata":{"env":"prod","env_id":"0","env_name":"","env_tag":"","protocol":"grpc","version":"","weight":"100"},"endpoints":["grpc://golden-us04.domob.cn:9789"]},{"id":"d742c26c247bbc377f8a99ac9c10e402f04da209","name":"bv_media_server.grpc","version":"","metadata":{"env":"prod","env_id":"0","env_name":"","env_tag":"","protocol":"grpc","version":"","weight":"100"},"endpoints":["grpc://golden-us03.domob.cn:9789"]}]
fatal error: concurrent map iteration and map write
#### Anything else we need to know?:
goroutine 170 [running]:
github.com/go-kratos/kratos/v2/registry.(*ServiceInstance).Equal(0xc00089f630, {0x1636220?, 0xc00089f630})
	/home/gitlab-runner/go/pkg/mod/github.com/go-kratos/kratos/v2@v2.7.3/registry/registry.go:88 +0x225
google.golang.org/grpc/attributes.(*Attributes).Equal(0x16092a0?, 0xc00067c260)
	/home/gitlab-runner/go/pkg/mod/google.golang.org/grpc@v1.63.2/attributes/attributes.go:96 +0x1b1
google.golang.org/grpc/resolver.addressMapEntryList.find(...)
[...]

PR #3386, opened 2024-07-31
fix(registry/polaris): fix concurrent map iteration and map write #3361

Ccheers:

fix concurrent map iteration and map write

#3361

### DIFF
--- a/contrib/polaris/registry.go
+++ b/contrib/polaris/registry.go
@@ -138,7 +138,7 @@ func (r *Registry) Register(_ context.Context, instance *registry.ServiceInstanc
 					Weight:       &weight,
 					Priority:     &r.opt.Priority,
 					Version:      &instance.Version,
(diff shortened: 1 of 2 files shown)


shenqidebaozi (dismissed):

contrib/polaris/registry.go
```diff
@@ -115,15 +115,16 @@ func (r *Registry) Register(_ context.Context, instance *registry.ServiceInstanc
 		}
 
 		// metadata
-		if instance.Metadata == nil {
-			instance.Metadata = make(map[string]string)
+		rmd := mapClone(instance.Metadata)
```
demoManito, line 118: image
this should also be changed to `rmd`?

demoManito (approved):

shenqidebaozi (approved):

PR #3416, opened 2024-09-05
chore(middleware/tracing): upgrade otel semconv

flc1125:

Upgrade otel semconv

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.24.0/specification/trace/semantic_conventions/http.md

### DIFF
--- a/middleware/tracing/span.go
+++ b/middleware/tracing/span.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/go-kratos/kratos/v2/metadata"
@@ -11,7 +12,7 @@ import (
 	"github.com/go-kratos/kratos/v2/transport/http"
 
 	"go.opentelemetry.io/otel/attribute"
-	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.24.0"
 	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc/peer"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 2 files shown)


middleware/tracing/span.go
```diff
@@ -31,12 +32,16 @@ func setClientSpan(ctx context.Context, span trace.Span, m interface{}) {
...
+				attrs = append(attrs,
+					semconv.HTTPRequestMethodKey.String(ht.Request().Method),
+					semconv.HTTPRouteKey.String(ht.PathTemplate()),
+					semconv.URLFragment(ht.Request().URL.Fragment),
+					semconv.URLPath(ht.Request().URL.Path),
+					semconv.URLFull(ht.Request().URL.String()),
+					semconv.URLQuery(ht.Request().URL.RawQuery),
+					semconv.URLScheme(ht.Request().URL.Scheme),
+					semconv.UserAgentOriginal(ht.Request().UserAgent()),
+				)
```
shenqidebaozi, line 44: It seems that there is no guarantee of forward compatibility

flc1125: Maintain the original attributes to ensure backward compatibility, but also add other attributes at the same time.

flc1125:
或者整个v2版本：`github.com/go-kratos/kratos/v2/middleware/tracing/v2`，增加一些自定义的设定span补充的逻辑？

shenqidebaozi:
There is no need to add V2 version

shenqidebaozi:
Retention is only to avoid affecting the query configuration that users have already used

flc1125:
> Retention is only to avoid affecting the query configuration that users have already used

OK。后面我在研究看看，如何在当前版本下，支持增加自定义处理 `Handler` 的支持。接口按如下定义OK没：

```go
type Handler interface {
	Start(ctx context.Context, req interface{}, span trace.Span) context.Context
	End(ctx context.Context, span trace.Span, reply interface{}, err error) context.Context
}
```

PR #3437, opened 2024-10-12
*: upgrade go1.22

hawkingrei:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/go-kratos/kratos/v2
 
-go 1.20
+go 1.22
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   build:
     strategy:
       matrix:
-        go: [ 1.20.x, 1.21.x, 1.22.x ]
+        go: [ 1.22.x, 1.22.x ]
```
shenqidebaozi, line 16: this needs to be modified

PR #3449, opened 2024-10-24
fix: unit test Recycle watch resolver tick goroutine

harbourlga:

The previous unit test did not recycle the watchresolved tick go routine. It is used to get the latest service information regularly. When running multiple unit tests, the v1/health/service/ interface of consul will be called concurrently, resulting in the X-Consul-Index returned by the server is not the expected result.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -25,7 +25,7 @@ jobs:
           ETCD_LISTEN_CLIENT_URLS: http://0.0.0.0:2379
           ETCD_ADVERTISE_CLIENT_URLS: http://0.0.0.0:2379
       consul:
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           ETCD_LISTEN_CLIENT_URLS: http://0.0.0.0:2379
           ETCD_ADVERTISE_CLIENT_URLS: http://0.0.0.0:2379
       consul:
-        image: consul:1.12.3
+        image: consul:1.13.9
```
shenqidebaozi, line 28: latest?

shenqidebaozi (approved):

PR #3481, opened 2024-12-05
fix: check server start error

### LINKED ISSUE #3480: Server start may fail and cause Register error
#### What happened:
 If server failed to start, it will cause Register fail too.

Return server error, instead register error if server error happens.
`{"level":"warn","ts":"2024-12-05T19:02:16.607470+0800","logger":"etcd-client","caller":"v3@v3.5.17/retry_interceptor.go:63","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc0003e4f00/127.0.0.1:2379","attempt":0,"error":"rpc error: code = Canceled desc = context canceled"}`

#### How to reproduce it (as minimally and precisely as possible):
Start a program which listen port A, use kratos to listen port A too.

The ctx will be cancled since the latst error, you guys did't handle it.

#### Environment:
- Kratos version (use `kratos -v`): v2.8.2
- Go version (use `go version`): go version go1.23.4 linux/amd64
- OS (e.g: `cat /etc/os-release`):
```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo

```
- Others:
None

SuzukiHonoka:

If server failed to start, it will cause Register fail too, and you can only get the error message related about Register, this is really frustrating.

fixes #3480

### DIFF
--- a/app.go
+++ b/app.go
@@ -3,6 +3,7 @@ package kratos
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
 	"os/signal"
 	"sync"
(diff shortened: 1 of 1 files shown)


app.go
```diff
@@ -108,11 +109,15 @@ func (a *App) Run() error {
 		})
 		wg.Add(1)
 		eg.Go(func() error {
-			wg.Done() // here is to ensure server start has begun running before register, so defer is not needed
+			defer wg.Done() // should call Done after Start since we need to capture any potential error
```
shenqidebaozi, line 112: Will the defer statement be executed when the server is blocked?

PR #3550, opened 2025-02-12
feat: disable grpc reflection configuration

BandlSkyler:
For security reasons, grpc reflection should not be registered directly by default

### DIFF
--- a/transport/grpc/server.go
+++ b/transport/grpc/server.go
@@ -113,6 +113,13 @@ func StreamInterceptor(in ...grpc.StreamServerInterceptor) ServerOption {
 	}
 }
 
+// DisableReflection disable grpc reflection.
+func DisableReflection() ServerOption {
+	return func(s *Server) {
+		s.disableReflection = true
+	}
+}
+
 // Options with grpc options.
 func Options(opts ...grpc.ServerOption) ServerOption {
 	return func(s *Server) {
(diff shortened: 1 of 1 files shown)


shenqidebaozi:
This change should not be made as it affects forward compatibility. It should be registered by default and optional not to register

BandlSkyler:
> This change should not be made as it affects forward compatibility. It should be registered by default and optional not to register

I see, but we should also ensure that users are not unknowingly enabling this feature and potentially exposing themselves to security risks without their consent.

transport/grpc/server.go
```diff
@@ -113,6 +113,13 @@ func StreamInterceptor(in ...grpc.StreamServerInterceptor) ServerOption {
 	}
 }
 
+// DisuseReflection disuse reflection.
+func DisuseReflection(enable bool) ServerOption {
```
shenqidebaozi, line 117: Why do we still need to pass the 'enable' parameter when the method is called 'Disuse'. Also, usually `disable` is used

transport/grpc/server.go
```diff
@@ -140,9 +147,11 @@ type Server struct {
 	customHealth     bool
 	metadata         *apimd.Server
 	adminClean       func()
+	disuseReflection bool
 }
 
 // NewServer creates a gRPC server by options.
+// By default, it will register gRPC reflection and gRPC admin.
```
shenqidebaozi, line 154: This annotation should not be added here

transport/grpc/server.go
BandlSkyler: yes, it's my mistake

transport/grpc/server.go
BandlSkyler: I'm not sure where to load annotations properly, please guide me.

BandlSkyler: I think we should add an annotation in kratos-layout about that grpc admin and grpc reflection are registered.

shenqidebaozi (approved):

PR #3552, opened 2025-02-18
chore: use the upper-level passed ctx in Discovery.cancel

1911860538:


demoManito (approved):

contrib/registry/discovery/impl_registrar.go
```diff
@@ -52,7 +52,7 @@ func (d *Discovery) Register(ctx context.Context, service *registry.ServiceInsta
 			case <-ticker.C:
 				_ = d.renew(ctx, ins)
 			case <-ctx.Done():
-				_ = d.cancel(ins)
+				_ = d.cancel(ctx, ins)
```
1911860538, line 55: Because of ctx.Done(), ctx might have already been canceled at this point. Passing ctx into the cancel method, where d.httpClient.R().SetContext(ctx) is called, could cause a failure. My modification might introduce a bug.

PR #3555, opened 2025-02-25
feat: http transport support get http.ResponseWriter

yuzilog:

In some special scenarios, such as the return of SSE messages, a little more flexibility can be added.


yuzilog:
> What is the usage scenario for this

For example, when developing AI-related applications, we can use streaming responses, or customize the format processing independent of the proto-based format, etc.

tonybase:
refer: https://github.com/go-kratos/kratos/pull/3624

PR #3667, opened 2025-05-16
fix(all): replace strconv.Itoa with strconv.FormatInt for int64 handling

1911860538:
Replaced instances of strconv.Itoa(int(a)) with strconv.FormatInt(a, 10) to safely handle int64 values. 
Using strconv.Itoa with int64 risks overflow on 32-bit systems when the value exceeds the maximum range of int32.  
Although 32-bit operating systems are rare in modern devices.


kagaya85 (approved):
LGTM

PR #3690, opened 2025-07-22
fix stale entries before the each pick operation

akoserwal:
The original implementation had a critical flaw in the `currentWeight` map management:
* No Cleanup Logic: The `currentWeight` map accumulated entries for every node address that was ever encountered
* Stale Entries Persist: When service discovery updated the node list (nodes added/removed), old entries remained in the map
* Weight Pollution: Stale entries could have accumulated very high or very low weights, affecting future calculations


shenqidebaozi (approved):

shenqidebaozi:
@tonybase PTAL

shenqidebaozi (approved):

shenqidebaozi:
@Windfarer @tonybase

PR #3698, opened 2025-08-14
refactor(consul)!: optimize ServiceInstance tags handling with registrarTags option

lin-snow:
## What this PR does / why we need it
Optimizes tag handling for the Consul registry:

- Introduces a new `registrarTags` option, allowing users to explicitly define the tags sent during service registration.  
- Adds the `Tags` field to `ServiceInstance`, enabling consistent retrieval on the discovery side.  
- Maintains backward compatibility.


lin-snow:
#3707

Issue #3219, opened 2024-03-02
[Feature] consule register　need  suport tags[] 


techidea8:
when we use traefik  as gateway 
and  consule as registry 
the tags is very important
but now  kratos just surport tag  `version` 
please suport more attribute as follow
```
// AgentServiceRegistration is used to register a new service
type AgentServiceRegistration struct {
	Kind              ServiceKind               `json:",omitempty"`
	ID                string                    `json:",omitempty"`
	Name              string                    `json:",omitempty"`
	Tags              []string                  `json:",omitempty"`
	Port              int                       `json:",omitempty"`
	Address           string                    `json:",omitempty"`
	SocketPath        string                    `json:",omitempty"`
	TaggedAddresses   map[string]ServiceAddress `json:",omitempty"`
	EnableTagOverride bool                      `json:",omitempty"`
	Meta              map[string]string         `json:",omitempty"`
	Weights           *AgentWeights             `json:",omitempty"`
	Check             *AgentServiceCheck
	Checks            AgentServiceChecks
	Proxy             *AgentServiceConnectProxyConfig `json:",omitempty"`
	Connect           *AgentServiceConnect            `json:",omitempty"`
	Namespace         string                          `json:",omitempty" bexpr:"-" hash:"ignore"`
	Partition         string                          `json:",omitempty" bexpr:"-" hash:"ignore"`
	Locality          *Locality                       `json:",omitempty" bexpr:"-" hash:"ignore"`
}

```

techidea8:
I understand your mean
in fact i support this feature using ` "github.com/hashicorp/consul/api" ` 
The best way is the team of go-kratos  modify kratos to support this feature
we work use devops platform like `codeup` 
it‘s not a good idea  for me to  modify kratos framework  
it  will affect our other application

PR #3707, opened 2025-08-27
feat(consul): support custom tags for service registration

HoronLee:
This change allows setting custom tags for a service registered in Consul by using the 'tags' key within the ServiceInstance's Metadata. The value should be a comma-separated string of tags.

这个PR解决了使用consul服务发现的时候无法为注册的服务添加自定义Tag的问题
---
This PR resolves an issue where custom tags could not be added to services registered with Consul during service discovery.

来自此Issue：#3219 
假设有这么一个应用场景：你的程序在docker、k8s等集群中，使用了traefik作为暴露服务和入口网关，并且你的微服务采用consul作为服务发现程序。这个时候可以选择使用traefik使用consulCatlog作为provider，前提是注册到consul的服务需要带有traefik所能识别的tag（而不是Metadata），可惜的是当前kratos注册到consul的服务并不能自定义tag
---
From this Issue: #3219. Let's assume an application scenario: your program is running in a cluster like Docker or Kubernetes, using Traefik as the service exposure and ingress gateway, and your microservices use Consul for service discovery. In this situation, you can opt to use Traefik with Consul Catalog as its provider. However, a prerequisite is that the services registered with Consul must have tags recognizable by Traefik (specifically tags, not Metadata). Unfortunately, services currently registered with Consul by Kratos do not support custom tags.

我在本地测试过了本功能，使用如下代码即可利用kratos原生的metadata为consul添加tag，使用","来进行tag的拆分
---
I have tested this feature locally. By using the following code, you can leverage Kratos's native metadata to add tags to Consul, using a comma (",") to split the tags.

```go
func newApp(logger log.Logger, reg registry.Registrar, gs *grpc.Server, hs *http.Server) *kratos.App {
	return kratos.New(
		kratos.ID(id),
		kratos.Name(Name),
		kratos.Version(Version),
		kratos.Metadata(map[string]string{
			"tags": "traefik.enable=true,traefik.http.routers.krathub-router.rule=Host(`krathub.r430.com`),traefik.http.services.krathub-svc.loadbalancer.server.port=8000",
		}),
		kratos.Logger(logger),
		kratos.Server(
			gs,
			hs,
		),
		kratos.Registrar(reg),
	)
}
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,3 +44,5 @@ require (
 	google.golang.org/genproto v0.0.0-20231212172506-995d672761c0 // indirect
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 )
+
+retract v2.9.0
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
 	"time"
 
 	"github.com/go-kratos/kratos/v2/encoding"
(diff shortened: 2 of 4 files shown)


longXboy:
lgtm

longXboy (dismissed):

shenqidebaozi:
This is an implicit logic

HoronLee:
We now configure Tags using the previously discussed Option pattern instead of the implicit logic. The example code is as follows:
```go
r := consul.New(client, consul.WithHealthCheck(true), consul.WithTags([]string{
        "traefik.enable=true",
        "traefik.http.routers.krathub-router.rule=Host(`krathub.r430.com`)",
        "traefik.http.services.krathub-svc.loadbalancer.server.port=8000",
    }))
```

PR #3755, opened 2025-11-28
feat(config): add option to control debug log printing for loaded con…

### LINKED ISSUE #3754: [Feature] Add a switch to support the debug of the "config loaded" information during configuration loading and enable the option to freely turn it off.
```go
// config.go 102
func (c *config) Load() error {
	for _, src := range c.opts.sources {
		kvs, err := src.Load()
		if err != nil {
			return err
		}
		// Add a flexible control switch to enable users to freely activate the "loaded" log, rather than relying on the log level for control.
		if c.printLoadedDebugLog {
			for _, v := range kvs {
				log.Debugf("config loaded: %s format: %s", v.Key, v.Format)
			}
		}
		
		if err = c.reader.Merge(kvs...); err != nil {
			log.Errorf("failed to merge config source: %v", err)
			return err
		}
		w, err := src.Watch()
		if err != nil {
			log.Errorf("failed to watch config source: %v", err)
			return err
		}
		c.watchers = append(c.watchers, w)
		go c.watch(w)
	}
	if err := c.reader.Resolve(); err != nil {
		log.Errorf("failed to resolve config source: %v", err)
		return err
	}
	return nil
}
```

aide-cloud:
feat(config): add option to control debug log printing for loaded configs

Adds a new `WithPrintLoadedDebugLog` option to allow users to enable or disable the debug logging of loaded configuration keys. This option defaults to false, ensuring no additional logs are printed unless explicitly enabled. The change also includes corresponding tests and updates to the config loading logic.

Close #3754

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -49,6 +49,7 @@ func New(opts ...Option) Config {
 		merge: func(dst, src any) error {
 			return mergo.Map(dst, src, mergo.WithOverride)
 		},
+		printLoadedDebugLog: true,
 	}
 	for _, opt := range opts {
 		opt(&o)
--- a/config/options.go
+++ b/config/options.go
@@ -37,6 +37,8 @@ func WithSource(s ...Source) Option {
 }
 
 // WithPrintLoadedDebugLog with config print loaded debug log.
+// Default is true.
+// When printLoadedDebugLog is set to false, the loaded configuration will not be printed in the log.
 func WithPrintLoadedDebugLog(printLoadedDebugLog bool) Option {
 	return func(o *options) {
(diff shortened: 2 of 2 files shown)


aide-cloud:
There is a scenario where the logs need to be initialized based on the configuration content. In this case, managing log printing during configuration loading by controlling the log level will become ineffective.

aide-cloud:
We do not want the loaded configuration information to be exposed in the console. Therefore, we prefer it to have an explicit switch to control it. This PR retains the original default scenario and does not cause any logical damage to the previous version.

PR #3767, opened 2025-12-03
fix(metadata): ensure Clone creates deep copy of slices

1911860538:
Previously, the Clone method was directly assigning slice values, which resulted
in sharing the underlying array between the original and cloned Metadata. This
could lead to unexpected side effects when modifying either the original or
cloned Metadata.


kagaya85 (approved):
LGTM

PR #3769, opened 2025-12-04
perf(config): move placeholder regex to package-level variable

1911860538:
Move the placeholder regex to package-level variable to avoid recompiling it on every expand() call, improving performance.


kagaya85 (approved):
LGTM

PR #3771, opened 2025-12-05
fix(contrib/registry/eureka): improve server picking and retry logic

1911860538:
This PR fixes potential issues in the Eureka client's server picking and retry logic.
`e.maxRetry` is not necessarily equal to `len(e.urls)`, 
because users can customize `e.maxRetry` using the `WithMaxRetry` function. 
When users set e.maxRetry=0, the for-loop in the `Client.do` method will not execute. 
When `e.maxRetry` is greater than `len(e.urls)`, array index out of bounds may occur. 
For example, if a user sets e.maxRetry=10 and len(e.urls)=2, 
when `pickServer` is executed with currentTimes=2, 
currentTimes%e.maxRetry=2, resulting in an array index out of bounds.

### DIFF
--- a/contrib/registry/eureka/client.go
+++ b/contrib/registry/eureka/client.go
@@ -288,7 +288,11 @@ func (e *Client) filterUp(apps ...Application) (res []Instance) {
 }
 
 func (e *Client) pickServer(currentTimes int) string {
-	return e.urls[currentTimes%e.maxRetry]
+	count := len(e.urls)
+	if count == 0 {
+		return ""
+	}
+	return e.urls[currentTimes%count]
 }
 
 func (e *Client) shuffle() {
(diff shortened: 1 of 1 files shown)


1911860538:
On the main branch, the following code can reproduce the array out-of-bounds issue.
### kratos/contrib/registry/eureka/client_test.go
```go
package eureka

import (
	"context"
	"testing"
)

// This test demonstrates an out-of-bounds panic in pickServer
// when maxRetry > len(urls).
func TestPickServer_OutOfBoundsPanic(t *testing.T) {
	// len(urls) = 2
	e := NewClient(
		[]string{
			"http://127.0.0.1:8080",
			"http://127.0.0.1:8081",
		},
		WithMaxRetry(3), // > len(urls)
	)

	// This call will eventually invoke pickServer with currentTimes = 2,
	// which leads to e.urls[2%3] == e.urls[2] and panics:
	// runtime error: index out of range [2] with length 2
	_ = e.Out(context.Background(), "app", "instance")
}

```
截屏2025-12-17 19 53 41

PR #3772, opened 2025-12-05
fix(contrib/registry/eureka): use shared http.Transport

1911860538:
Use a shared http.Transport to enable connection reuse and prevent connection accumulation.
Previously, each NewClient() call create a new http.Transport, which led to:
- Independent connection pools per client instance
- Inability to reuse connections across clients
- Connection accumulation due to keep-alive connection not being shared  

In theory, NewClient should only be called once. However, users may call it multiple.
Meanwhile, to follow Go best practices, we should modify the code to reuse http.Transport.
Ref: https://github.com/golang/go/blob/master/src/net/http/client.go#L33


contrib/registry/eureka/client.go
```diff
@@ -129,18 +129,26 @@ type Client struct {
...
-func NewClient(urls []string, opts ...ClientOption) *Client {
-	tr := &http.Transport{
-		MaxIdleConns: maxIdleConns,
-	}
+var clientTransport = &http.Transport{
+	DialContext: (&net.Dialer{
+		Timeout:   5 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}).DialContext,
+	IdleConnTimeout:       30 * time.Second,
```
tonybase: It should be set to 90 seconds, which needs to be greater than the connection KeepAlive time.

tonybase (approved):

Issue #3820, opened 2026-05-10
[Proposal] Kratos v3: Clean up historical design issues and reduce external dependencies


tonybase:
## Proposal Description

Kratos v3 is intended to clean up historical design issues accumulated in v2, reduce coupling between the framework and specific external dependencies, and align with newer standards in the Go ecosystem. This proposal contains several v3 cleanup improvements, including a small number of breaking changes.

## TODOs

- [x] Split `encoding/json` into `encoding/json` and `encoding/protojson` (PR #3823)
- [x] Replace the log package with a lightweight standard-library `slog` facade and compose OTEL through contrib (PR #3821)
- [x] Move `middleware/auth/jwt` to contrib (PR #3824)
- [x] Remove the `aegis` dependency and move the default circuit breaker/rate limiter implementations into internal packages (PR #3826)
- [x] Evaluate removing the `api` package (PR #3825)
- [x] HTTP router and proto gen http refactor conclusion: v3 will continue to keep `github.com/gorilla/mux`, remove the `transport/http/binding` package, switch generated code to `http.BuildPath`, and use `application/protojson` by default through `Accept` / `Content-Type` (PR #3827)
- [x] Clean up config package dependencies: remove `dario.cat/mergo` and keep `github.com/fsnotify/fsnotify` because the file source still requires it (PR #3828)
- [x] Add protoc-gen-go-http streaming support with SSE, WebSocket, HttpBody upload/download, and Accept/Content-Type codec negotiation (PR #3829)

## V3 Integration Branch

- Unified integration PR/branch: PR #3822 (`v3` -> `main`)
- Future v3 changes should first merge into the `v3` branch, then be collected into the main line through #3822.

---

### Problem

The current `encoding/json` package mixes standard JSON encoding/decoding and Protobuf JSON encoding/decoding in the same codec, both registered under the `"json"` name:

```go
// encoding/json/json.go
func (codec) Marshal(v any) ([]byte, error) {
    switch m := v.(type) {
    case json.Marshaler:        // Standard JSON
        return m.MarshalJSON()
    case proto.Message:         // Protobuf JSON
        return MarshalOptions.Marshal(m)
    default:                    // Standard JSON
        return json.Marshal(m)
    }
}
```

This design causes several issues:

- Users cannot explicitly choose the encoding strategy they want: standard JSON or proto JSON.
- Protojson dependencies (`google.golang.org/protobuf/encoding/protojson`) are always pulled in, even when a project does not use protobuf.
[...]

PR #3822, opened 2026-05-12
break: update kratos modules to v3

tonybase:

- Update the root module from `github.com/go-kratos/kratos/v2` to `github.com/go-kratos/kratos/v3`.
- Update `cmd` and `contrib` module paths, imports, docs, and generated references from `/v2` to `/v3`, including `cmd/protoc-gen-go-errors/v3`.
- Update internal module `require` and `replace` references to `github.com/go-kratos/kratos/v3 v3.0.0`.
- Update golangci-lint to `v2.12.2` and fix new lint findings.

Refs #3820.

This PR owns the shared `v3` integration branch; follow-up v3 feature PRs should target `v3` first.

Local validation:
- `make clean`
- `make lint`
- `go build ./...`
- `go test ./...`
- `go test -race ./...`
- `go build ./...` in `cmd/kratos`
- `go test ./...` in `cmd/protoc-gen-go-errors`
- `go test ./...` in `cmd/protoc-gen-go-http`

Service-backed contrib tests are covered by GitHub Actions, which starts etcd, Consul, Nacos, and Polaris.


transport/http/stream.go
```diff
@@ -0,0 +1,704 @@
...
+	}
+	prepareClientRequest(client, req, contentType, c)
+	ctx = transport.NewClientContext(ctx, &Transport{
+		endpoint:     client.opts.endpoint,
+		reqHeader:    headerCarrier(req.Header),
+		operation:    c.operation,
+		request:      req,
+		pathTemplate: c.pathTemplate,
+	})
+	res, err := client.do(req.WithContext(ctx)) //nolint:bodyclose // newSSEClientStream owns and closes res.Body on success.
```
shenqidebaozi: It seems that middleware is not supported

transport/http/stream.go
```diff
@@ -0,0 +1,704 @@
...
+	}
+	return codec.Unmarshal(data, v)
+}
+
+func readWebSocketMessage(conn *websocket.Conn, m any, codec encoding.Codec) error {
+	for {
+		messageType, data, err := conn.ReadMessage()
+		if err != nil {
+			if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
+				return io.EOF
```
shenqidebaozi, line 728: Should the normal shutdown of `io.eof` be considered as successful? Otherwise, it will be handled by the selector `DoneInfo.Err`, affecting the load balancer strategy

transport/http/stream.go
```diff
@@ -0,0 +1,704 @@
...
+
+func (s *sseClientStream) CloseAndRecv(any) error {
+	return stderrors.New("SSE client stream does not support CloseAndRecv")
+}
+
+func (s *sseClientStream) SendMsg(any) error {
+	return stderrors.New("SSE client stream does not support SendMsg")
+}
+
+func (s *sseClientStream) RecvMsg(m any) error {
```
shenqidebaozi, line 320: Do we need to actively close the body when receiving eof

cmd/protoc-gen-go-http/httpTemplate.tpl
```diff
@@ -69,16 +164,172 @@ func New{{.ServiceType}}HTTPClient (client *http.Client) {{.ServiceType}}HTTPCli
...
+func (c *{{$svrType}}HTTPClientImpl) {{.Name}}(ctx context.Context, opts ...http.CallOption) ({{$svrType}}_{{.Name}}Client, error) {
+	pattern := "{{.PathTemplate}}"
+	opts = append([]http.CallOption{
+		http.Accept("application/protojson"),
+		{{- if not .BodyHTTPBody}}
+		http.ContentType("application/protojson"),
+		{{- end}}
+		http.Operation(Operation{{$svrType}}{{.OriginalName}}),
+		http.PathTemplate(pattern),
+	}, opts...)
```
shenqidebaozi, line 267: Generated HTTP clients default to protojson, but the codec may not be registered

  cmd/protoc-gen-go-http/httpTemplate.tpl:260-264 generates clients with Accept("application/protojson") / ContentType("application/protojson"). The
  generated file only imports transport/http (internal/testdata/helloworld/helloworld_http.pb.go:7-10), so encoding/protojson is not guaranteed to be
  registered.

  Then transport/http/client.go:351 and transport/http/codec.go:115-116 call Marshal on encoding.GetCodec(name) without checking nil. If an app only
  imports the generated HTTP client/server code and does not separately blank-import github.com/go-kratos/kratos/v3/encoding/protojson, this can pan
  ic.

  Suggested fix: either blank-import the default codec inside transport/http, or make generated HTTP code blank-import encoding/protojson. Also add
  nil checks around GetCodec to return a clear error instead of panicking.

transport/http/stream.go
```diff
@@ -0,0 +1,763 @@
...
+
+func (s *websocketClientStream) Header() (metadata.MD, error) {
+	return metadataFromHeader(s.header), nil
+}
+
+func (s *websocketClientStream) Trailer() metadata.MD {
+	return nil
+}
+
+func (s *websocketClientStream) CloseSend() error {
```
shenqidebaozi, line 407: WebSocket CloseSend does not prevent later Send calls

  transport/http/stream.go:403-407 sends the end control frame but does not mark the send side as closed. transport/http/stream.go:433-440 still
  allows later SendMsg calls.

  That violates gRPC-style stream semantics: after CloseSend, further sends should fail. It can also put the server in an odd state where it has
  already observed EOF but the client keeps writing messages.

  Suggested fix: track a sendClosed state separately from full stream close, guard it with a mutex or atomic, and make SendMsg return a clear error
  after CloseSend. Add a test for CloseSend followed by Send.

  I stopped dependency-related test work per your instruction. Static checks plus the tests that did not need those blocked dependencies showed cmd/
  protoc-gen-go-http, log, config, and base middleware passing.

shenqidebaozi:
1. WebSocket stream state handling in transport/http/stream.go
     Replace the single unsynchronized closed bool with clearer state:
      - sendClosed
      - closed
      - closeOnce sync.Once

     This avoids duplicate close behavior, data races, and makes CloseSend semantics closer to gRPC.
  2. Add nil protection around encoding.GetCodec(...)
     Several paths assume codecs always exist. After splitting json and protojson, that assumption is fragile. A small helper like codecOrError(name)
     would make failures explicit instead of panicking.
  3. Escape path variables in BuildPath
     transport/http/path.go inserts values directly into URL paths. Values containing spaces, ?, #, %, etc. can produce invalid or surprising URLs.
     Use url.PathEscape, but handle resource-name templates like {name=publishers/*/books/*} carefully because / may be intentional.
  4. Reduce runtime work in BuildPath
     It currently regex-parses templates and encodes all query values at runtime. Since most calls come from generated code, path variable names and
     omitted fields could be generated statically to reduce regex/map work. Lower priority, but cleaner and faster.
  5. Avoid the 1 MB SSE event limit
     bufio.Scanner in sseClientStream is capped at 1 MB. Large streaming messages will fail. bufio.Reader.ReadString('\n') would be more flexible, or
     the size limit should be documented.
  6. Clarify or implement streaming trailers
     serverStream.SetTrailer stores trailers, but SSE/WebSocket close paths do not send them. Either implement trailer delivery or document that HTTP
     streaming trailers are currently ignored.

shenqidebaozi (approved):

PR #3856, opened 2026-07-21
fix(transport/http): escape path parameters safely

### LINKED ISSUE #3853: bug: URL path parameters can break URL structure due to lack of partial escaping
#### Description

The `BuildPath` function constructs an HTTP request path from a path template and a request message (`msg`). However, values extracted from the message fields are injected directly into the path template without any escaping.

While certain path templates (such as resource name patterns like `{name=publishers/*/books/*}`) are designed to accept and preserve forward slashes (`/`), the complete lack of escaping allows characters like `?`, `#`, or spaces to break the URL structure.

#### Affected Code

The raw string returned from `queryParams.Get(key)` is substituted directly:
```transport/http/path.go```
```go
path = pathTemplateParamRE.ReplaceAllStringFunc(pathTemplate, func(in string) string {
    matches := pathTemplateParamRE.FindStringSubmatch(in)
    key := matches[1]
    pathParams[key] = struct{}{}
    return queryParams.Get(key) // <--- Problem: Injected directly without validation/escaping
})

```

#### Technical Impact

If the fields in the request message contain specific special characters, it could lead to:

* **URL Structure Corruption:** Spaces or control characters causing malformed HTTP requests.
* **Query/Fragment Injection:** An attacker or unexpected data injecting a `?` or `#` into the path segment, modifying query parameters or the URL fragment prematurely.
* **Unintended Routing:** Misrouting requests at the downstream service or reverse proxy level.

#### Suggested Fix Approach

Using standard `url.PathEscape` directly is **incorrect** because it breaks existing resource name patterns by escaping `/` into `%2F`.

Instead, the function should escape only the unsafe characters while preserving `/`. One common approach is to split the value by `/`, escape each segment individually using `url.PathEscape`, and then join them back together.

For example, a helper function could be used during substitution:

```go
// Helper to escape path segments while preserving slashes for resource names
func escapePathValue(v string) string {
    segments := strings.Split(v, "/")
    for i, s := range segments {
        segments[i] = url.PathEscape(s)
    }
    return strings.Join(segments, "/")
}

// Inside BuildPath:
path = pathTemplateParamRE.ReplaceAllStringFunc(pathTemplate, func(in string) string {
    matches := pathTemplateParamRE.FindStringSubmatch(in)
    key := matches[1]
    pathParams[key] = struct{}{}
    return escapePathValue(queryParams.Get(key)) // <--- Fix: Escape segments while keeping '/'
})

```

ZYZ666-RGB:

`BuildPath` previously substituted path parameter values without escaping them. Values containing spaces, `?`, or `#` could therefore corrupt the URL structure or inject query/fragment delimiters.

This change:

- Escapes each slash-separated path segment with `url.PathEscape`.
- Preserves `/` for resource names such as `publishers/go/books/kratos`.
- Preserves RFC 3986 sub-delimiters that were already accepted by `BuildPath`, avoiding compatibility regressions.
- Adds regression coverage for query/fragment characters, spaces, resource-name slashes, and valid path-segment characters.

Fixes #3853

Focused tests pass:

```text
go test ./transport/http -run '^TestBuildPath' -count=1
```


ZYZ666-RGB:
Hi maintainers, the **Non-English Comments Check** failed before the scanning step ran.

The failure occurred in `actions/checkout@v6`: it refused to check out fork code because this workflow uses `pull_request_target`, which runs in the trusted base-repository context. The subsequent non-English comment scan was skipped.

I also reproduced the workflow's grep check locally against the two changed files, and it found no non-English comments.

Failed run: https://github.com/go-kratos/kratos/actions/runs/29814185538

Could you please review the workflow configuration or re-run the appropriate check after it is adjusted? Thank you!

ZYZ666-RGB:
Hi maintainers, just following up on this PR. The Go and Lint checks are green, and the focused tests cover query/fragment characters, spaces, resource-name slashes, and preserved valid path-segment characters. Please let me know if there are any compatibility cases or changes you would like me to address. Thanks!

